Issue #22, opened 2018-12-20
Unit test for Contact us API endpoint


OtisRed:
@stanislawK

Issue #21, opened 2018-12-20
Py-test infrastructure


OtisRed:
@stanislawK discovered that there are some problems with writing tests that are related to the organization of files in backend folder - apparently there is some issue with "circular imports". We need to solve this in order to be able to write tests that work. 

### Until now there were two voices on the matter:
* Staszek suggested we might have to reorganize the structure of files 
* Jacek suggested that we need "factory" - he might have been thinking about [this](http://flask.pocoo.org/docs/1.0/patterns/appfactories/)?

stanislawK:
I think that we should create app factory (like in docks which you mentioned). Probably also setup.py like [there](http://flask.pocoo.org/docs/1.0/patterns/packages/). It requires different app's structure. I'm trying figure it out. We need something like that, more or less:
```

└── backend
    ├── api
    │   ├── config.py
    │   ├── __init__.py
    │   ├── mail.py
    │   ├── manage.py
    │   ├── migrations
    │   │   ├── ...
    │   ├── models.py
    │   ├── serializers
    │   │   ├── ...
    │   └── views.py
    ├── Dockerfile
    ├── setup.py
    └── tests
        ├── ...
```

PR #23, opened 2019-01-01
[17] Add contact us API endpoint

stanislawK:
In development and testing env sending emails is suppressed by set TESTING to True. For production providing valid data follow by ProductionConfig class in config.py is required. Also proper cfp email address has to be given in Message recipients.

Writing tests #22 requires additional steps: issue #21

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 node_modules/
 
 yarn-error.log
+
+__pycache__/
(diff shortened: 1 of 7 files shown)


OtisRed:
@stanislawK since you've pushed backend to the PR #24 can we close this PR? Or does it contain something that #24 does not?

PR #25, opened 2019-01-11
[WIP][21] pytest infrastructure

stanislawK:
New app structure proposal


stanislawK:
cdf team decided that we shouldn't use setup.py

PR #30, opened 2019-02-13
[21] Pytest infrastructure

stanislawK:
Added:
- app factory 
- separate file with config
- deleted manage.py, and used flask CLI instead


kristhina:
Worked (docker compose and pytest)

PR #31, opened 2019-02-13
[2][8]Login, Logout and JWT authorization

jacekkalbarczyk:
Added login and logout endpoints.Created new db table for JWT tokens blacklisting.


backend/blueprints/auth.py
```diff
@@ -0,0 +1,107 @@
...
+    if not request.is_json:
+        return jsonify({"msg": "No input data provided"}), 400
+
+    schema = LoginSchema()
+    try:
+        result = schema.load(request.json)
+    except ValidationError as error:
+        return jsonify({"msg": "Wrong input data",
+                        "errors": error.messages}), 400
+
```
stanislawK: It returns two separate error messages at once. Is it intended?
```
{
    "errors": {
        "username": [
            "Shorter than minimum length 5."
        ]
    },
    "msg": "Wrong input data"
}
```

backend/blueprints/auth.py
```diff
@@ -0,0 +1,107 @@
...
+        add_token_to_database(refresh_token, app.config['JWT_IDENTITY_CLAIM'])
+
+        ret = {
+            'access_token': access_token,
+            'refresh_token': refresh_token
+        }
+        return jsonify(ret), 201
+    else:
+        return jsonify({"msg": "Not authorized"}), 401
+
```
stanislawK: We maybe should create separate issue for refresh_token, and all related to it things like refreshing endpoint, and refresh logout?

PR #32, opened 2019-02-16
[17]add contact us endpoint

stanislawK:


OtisRed:
@stanislawK ping me when conflicts are resolved so it could be tested properly. I think if we merge it first then it would be easier to double check frontend.

PR #33, opened 2019-02-17
[26]tests for models and db

stanislawK:
To run tests:
- get into backend container: docker exec -ti codeforpoznanpl_v3_backend_1 bash
- type pytest and run


OtisRed:
@stanislawK ping me when conflicts are resolved for testing

backend/tests/test_db.py
```diff
@@ -0,0 +1,65 @@
...
+import pytest
+
+from backend.models import User, Participant, Hacknight
+
+
+def test_add_new_user_to_db(_db, new_user):
+    """
+    GIVEN mocked db session, and user data
+    WHEN a new user added to db
+    THEN check the user data from db
```
w1stler: Gherkin syntax is descriptive and used for the BDD approach. I would stick with classic docstring convention. Though we can discuss this matter.

backend/tests/conftest.py
```diff
@@ -28,3 +29,51 @@ def app():
...
+            'password': 'TestPassword'}
+    return user
+
+
+@pytest.fixture
+def new_participant():
+    participant = {'name': 'Jon',
+                   'lastname': 'Doe',
+                   'email': 'test@test.com',
+                   'phone': '123456789'}
```
w1stler: I don't like this indentation.

Issue #43, opened 2019-03-20
Populate database


stanislawK:
We should have some script which allow us to add data (some user, participant, and hacknight) to db in simple way

PR #45, opened 2019-04-17
[42] participant and hacknight model relationship 

kristhina:


backend/models.py
```diff
@@ -36,13 +36,28 @@ class Participant(db.Model):
     email = Column(String(200))
     github = Column(String(200), default="")
     phone = Column(String(13))
+    hacknights = db.relationship("Hacknight",
+                                 secondary=participant_hacknight,
+                                 lazy='subquery',
+                                 backref=db.backref('participants', lazy=True))
```
stanislawK: This looks like bidirectional many to many relationship. So either you should remove one relationship (from child class), or use back_populates instead of backref if you would like to keep this two directions.

backend/models.py
```diff
@@ -36,13 +36,28 @@ class Participant(db.Model):
...
+                                   secondary=participant_hacknight,
+                                   lazy='subquery',
+                                   backref=db.backref('hacknights', lazy=True))
+
+
+participant_hacknight = db.Table(
+    'participant_hacknight',
+    db.Column('participant_id', db.Integer, db.ForeignKey('participant.id')),
+    db.Column('hacknight.id', db.Integer, db.ForeignKey('hacknight.id'))
+)
```
stanislawK: Association table should be before participant and hacknight models

backend/models.py
```diff
@@ -43,6 +50,10 @@ class Hacknight(db.Model):
     __tablename__ = 'hacknight'
     id = Column(Integer, autoincrement=True, primary_key=True)
     date = Column(Date)
+    participants = db.relationship("Participant",
+                                   secondary=participant_hacknight,
+                                   lazy='subquery',
+                                   backref=db.backref('hacknights', lazy=True))
```
w1stler: I don't like this indentation.

stanislawK: I'm going to fix it in #43

PR #86, opened 2019-05-29
[16] footer component

Loczi94:


frontend/src/components/PageFooter.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+    <div>
```
Bosmanfrx: Why encapsulate footer in div and not just put in footer directly?

jacekkalbarczyk: That's a VueJS's requirement.

frontend/src/components/PageFooter.vue
```diff
@@ -0,0 +1,24 @@
...
+        <footer>Copyright © Code for Poznan 2019</footer>
+    </div>
+</template>
+
+<script>
+export default {
```
jacekkalbarczyk: I guess <script> section is not required if component doesn't contain JS logic. Or is it (for more elegant code styling)? To be considered.

frontend/src/components/PageFooter.vue
```diff
@@ -0,0 +1,24 @@
...
+
+<style lang="scss" scoped>
+@import "./../main.scss";
+footer {
+    color: $white;
+    background-color: $blue-nav-footer;
+    padding: 25px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
jacekkalbarczyk: I'm not sure if this line(align-items) is needed, it doesn't seem to do anything (correct me if I'm wrong).

Loczi94: It aligns items vertically, but in this case you're right - it's not needed.

PR #89, opened 2019-06-01
Updated lint config

adamgajzlerowicz:
This PR:

- extends linting configuration with extra plugins
- adds [prettier](https://prettier.io/docs/en/) for unified experience across projects
- add husky for pre push lint check, to ensure only healthy code in the main repo


frontend/.eslintrc.json
```diff
@@ -0,0 +1,30 @@
+{
+  "root": true,
+  "env": {
+    "node": true
+  },
+  "plugins": [
+    "prettier"
```
Bosmanfrx: I think, that prettier plugin to eslint is unnecessary. Prettier itself should be transparent to development process. It should be added on pre-commit hook instead.

adamgajzlerowicz: Current strategy uses prettier as a plugin, thus there is no need to use prettier directly as eslint uses it for linting. Prehook is also already added.

Bosmanfrx: Exactly, you need to run prettier yourself or manually adjust code. Why not just hook prettier to pre-commit to do it for you?

adamgajzlerowicz: it is already done.

Bosmanfrx: It is not. In pre-commit hook `yarn lint` is ran and not `yarn lint:fix`. Also changed files need to be restaged in commit if you decide to use `yarn lint:fix`. Also IMHO adding Prettier as plugin adds unnecessary information noise in IDE.

adamgajzlerowicz: Only check with no fix is used so you can stage only part of the changes. Making fix and staging files for devs will cause unwanted changes to go to repository

PR #73, opened 2019-05-22
Updated project structure to separate css from html and js

adamgajzlerowicz:

PR #92, opened 2019-06-05
[15] social media component

Loczi94:


frontend/src/components/SocialMedia.vue
```diff
@@ -0,0 +1,82 @@
...
+        <h2>MEDIA</h2>
+        <ul>
+            <li v-for="item in media" :key="item.link">
+                <a :href="item.link" target="_blank" class="circle">
+                    <v-icon> {{item.icon}} </v-icon>
+                </a>
+            </li>
+        </ul>
+    </div>
+</template>
```
OtisRed: Please look up the vue component structure proposed in #73. The main idea behind it is to avoid scrolling between logic and template which can get a real pain in the bottom when the .vue file gets longer. Accordingly template should be separate .html file imported in .vue file.

frontend/src/components/SocialMedia.vue
```diff
@@ -0,0 +1,82 @@
...
+                    link: 'https://github.com/CodeForPoznan',
+                },
+                {
+                    icon: 'fab fa-linkedin-in',
+                    link: 'https://www.linkedin.com/company/codeforpoznan/',
+                },
+                {
+                    icon: 'fab fa-facebook-f',
+                    link: 'https://www.facebook.com/CodeForPL/',
+                },
```
OtisRed: Previously there was also gitter icon here and as we discussed last week since we don't use we shall skip it then. But we also found out that there is an easy way to open the Slack workspace so you might add the icon for Slack and referenced it with this link: https://join.slack.com/t/codeforpoznan/shared_invite/enQtNjQ5MTU1MDI0NDA0LWJmMmM2NjI0MGNiZjJjMGM0NzMzOTEzOGY5YWZkZWNiMzZlZTM2ZGJlYzRkYTlmZTFlMTMzZjJkZTc5YjhhYjg (or some shortened version of it)

frontend/src/components/SocialMedia.vue
```diff
@@ -0,0 +1,82 @@
...
+        .v-icon {
+            color: $blue-media-hr;
+        }
+}
+.v-icon {
+    color: $white;
+    padding: 10px;
+    border-radius: 50%;
+}
+</style>
```
OtisRed: Accordingly to #73 style section should be a seperate file

frontend/src/components/SocialMedia.vue
Loczi94: I know the idea, but Adam said that for now we should stick to old project structure (html, js and scss in one file).

frontend/src/components/SocialMedia.vue
Loczi94: Okay, I'll do it :)

frontend/src/components/SocialMedia.vue
OtisRed: Outdated

PR #102, opened 2019-06-18
Add flask_restful endpoints (fixes #85)

### LINKED ISSUE #85: RESTing up endpoints
@stanislawK belives that it might be good idea to make our endpoints more RESTful

stanislawK:


backend/resources/auth.py
```diff
@@ -0,0 +1,102 @@
...
+    def delete(self):
+        jti = get_raw_jwt()['jti']
+        token = JWTToken.query.filter_by(jti=jti).one()
+        token.revoked = True
+        db.session.commit()
+        return {"msg": "Successfully logged out"}, 200
+
+
+def add_token_to_database(encoded_token, identity_claim):
+    """
```
w1stler: Why newline?

backend/resources/auth.py
```diff
@@ -0,0 +1,102 @@
...
+    Adds a new token to the database. It is not revoked when it is added.
+    :param identity_claim:
```
w1stler: Please add param description.

backend/resources/auth.py
```diff
@@ -0,0 +1,102 @@
...
+        user_identity=user_identity,
+        expires=expires,
+        revoked=revoked,
+    )
+    db.session.add(db_token)
+    db.session.commit()
+
+
+def is_token_revoked(decoded_token):
+    """
```
w1stler: Why newline?

backend/resources/contact.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class SendMessage(Resource):
+    def post(self):
+        try:
+            new_msg = message_schema.load(request.json)
+        except ValidationError as err:
+            return {"message": err.messages}, 400
+        msg = Message(
+            subject='Email z cdf_v3 od {}'.format(new_msg["name"]),
```
w1stler: cdf? :D

backend/resources/contact.py
```diff
@@ -0,0 +1,29 @@
...
+            sender=new_msg["email"],
+            reply_to=new_msg["email"],
+            recipients=["cfp_v3@cfp_v3.com"]  # has to be change to valid email
```
w1stler: I would stick with test domain (we don't want to actually send emails when runnning tests). I would go with something like: cfp@codeforpoznan.test. Also typo in comment. ;]

backend/resources/auth.py
stanislawK: I tried to follow PEP8: "Surround top-level function and class definitions with two blank lines."

backend/resources/auth.py
stanislawK: I tried to follow PEP8: "Surround top-level function and class definitions with two blank lines."

backend/resources/auth.py
stanislawK: I will have to talk with @jacekkalbarczyk because I simply copied his feature and refactored it due to flask RESTful requirements

backend/resources/auth.py
w1stler: I was refering to the docstring.

Please do:

```def foo_bar():
    """This is test description."""

File: backend/resources/auth.py (hunk 3 above)
Comment (w1stler) [reply backend/resources/auth.py]: ditto

File: backend/resources/auth.py (hunk 1 above)
Comment (magul) [reply backend/resources/auth.py]: @stanislawK https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

File: backend/tests/test_contact_us.py [hunk 6]
```diff
@@ -20,7 +20,7 @@ def test_contact_us_without_email(client, new_msg):
     """Test sending message with no email provided."""
     del new_msg["email"]
     rv = client.post('/send-email/', json=new_msg)
-    response = rv.get_json()
+    response = rv.get_json()['message']
```
magul: Two remarks here:
 * there's something essentially wrong when we're returning something like that:
```
{
    'message': {
        'content': {
            'code': 400,
            'message': 'Content of message is required'
        }
    }
}

```
as the body of a response. That need to be discussed so created separate ticket for that: https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/132
 * if you have `response["content"]["message"]` at the bottom I don't see a lot of reason to do something like that: `response = rv.get_json()['message']`. Instead, you can have something like:
```
    response = rv.get_json()
    assert rv.status == "400 BAD REQUEST"
    assert response['message']["content"]["message"] == "Content of message is required"

```

magul: Also we could use here `asser rv.status_code == http.HTTPStatus.400_BAD_REQUEST`

jacekkalbarczyk:
Changes requested by @w1stler regarding parameters description and docstrings are to be resolved in issue #138.

w1stler:
Approving, as other tickets are created.

PR #105, opened 2019-06-19
Tests for get participants endpoint (fixes #104)

### LINKED ISSUE #104: Get participants list enpoint test

stanislawK:


backend/tests/test_participants_list.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_get_participants_with_empty_db(client, access_token):
+    """Test get participants list when no participant in db."""
+    rv = client.get(
+        '/participants',
+        headers={'Authorization': 'Bearer {}'.format(access_token)}
+    )
+    response = rv.get_json()
+    assert rv.status == "404 NOT FOUND"
```
w1stler: Shouldn't that be 200 with 0 participants?

kristhina: In my opinion there should not occur a situation where there are 0 participants.
At the very beginning all Code of Poznań users should be added as participant and participants_list should return all of them.

w1stler: If you can do something with the application, someone will do that. ;-)

magul: It should be `200 OK` here. That's a bug in the application and I created a separate issue to solve that: #130

backend/tests/conftest.py
```diff
@@ -117,3 +118,11 @@ def new_msg():
         "content": "Lorem Ipsum cos tam cos tam"
     }
     return msg
+
+
+@pytest.fixture
+def add_participants(app, _db):
+    db = _db
```
magul: Do not see a lot of reason for such assignment. Would say I'm fine with overwriting that symbol (just as `def add_participants(app, db)` or calling it at the end as `_db.session.commit()`.

backend/tests/test_participants_list.py
```diff
@@ -0,0 +1,35 @@
+import pytest
```
magul: unused import

backend/tests/test_participants_list.py
```diff
@@ -0,0 +1,35 @@
...
+def test_get_participants_when_logged_in(
+        client, access_token, add_participants
+     ):
+    """Test get participants list for logged in user."""
+    rv = client.get(
+        '/participants',
+        headers={'Authorization': 'Bearer {}'.format(access_token)}
+    )
+    response = rv.get_json()
+    assert rv.status == "200 OK"
```
magul: I would check `rv.status_code == HTTPStatus.200_OK`

frontend/yarn.lock
```diff
@@ -9079,10 +9079,10 @@ vuelidate@0.7.4:
   resolved "https://registry.yarnpkg.com/vuelidate/-/vuelidate-0.7.4.tgz#5a0e54be09ac0192f1aa3387d74b92e0945bf8aa"
   integrity sha512-QHZWYOL325Zo+2K7VBNEJTZ496Kd8Z31p85aQJFldKudUUGBmgw4zu4ghl4CyqPwjRCmqZ9lDdx4FSdMnu4fGg==
 
-vuetify@1.4.0:
-  version "1.4.0"
-  resolved "https://registry.yarnpkg.com/vuetify/-/vuetify-1.4.0.tgz#ad5446a319cf4b090d3e14f9c3ee2ac53d545415"
-  integrity sha512-Q6b5VrkGIhPTN+r/uw6P5UTYVQy0h3d4vRW5363qHTgzQfyYQk8BDlJNCTJj0Zu/z3aBq79zbyF88jbLlV9vxQ==
+vuetify@1.5.14:
```
magul: change unconnected with a pull request.

backend/tests/test_participants_list.py
kristhina: As it was error in my endpoint (get participants list), I corrected ParticipantsList endpoint:
pull request #139: changed response and http code (fixes #130)

PR #114, opened 2019-06-26
Participant details endpoint (fixes# 110)

stanislawK:


backend/serializers/participant_serializer.py
```diff
@@ -3,14 +3,36 @@
...
+        )
+        dump_only = ('id', 'hacknights')
 
     name = fields.Str(required=True, validate=[validate.Length(max=50)])
     lastname = fields.Str(validate=[validate.Length(max=50)])
     email = fields.Email(validate=[validate.Length(max=200)])
     github = fields.Str(validate=[validate.Length(max=200)])
+    hacknights = fields.Nested(
+        'HacknightSchema',
+        exclude=('participants', ),
```
w1stler: That's just one parameter, comma could be removed?

stanislawK: That is tuple, and comma is required. We could pass list instead of tuple, where comma is not required, but tuple is more common in marshmallow docs.

backend/serializers/hacknight_serializer.py
```diff
@@ -0,0 +1,17 @@
...
+        dump_only = ('id',)
+
+    participants = fields.Nested(
+        'ParticipantSchema',
+        only=('id', 'github'),
+        many=True
+    )
+
+
+hacknight_schema = HacknightSchema()
```
magul: I personally don't see a lot of reason to have that:
```
hacknight_schema = HacknightSchema()
hacknights_schema = HacknightSchema(many=True)
```
 For a while, I was thinking that we have two separate serializers. Would prefer to use just `HacknightSchema` and `HacknightSchema(many=True)`.

backend/serializers/hacknight_serializer.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class HacknightSchema(Schema):
+    class Meta:
+        fields = ('id', 'date', 'participants')
+        dump_only = ('id',)
+
+    participants = fields.Nested(
+        'ParticipantSchema',
+        only=('id', 'github'),
```
magul: I would prefer to have it consistent (if we're using `exclude` in `hacknights` would prefer it also here).

Also would believe, that if we're going to return a list of entities (`many=True`) list of fields could be limited, as participants list should not return hacknights list for every participant and don't see a lot of reason to return all participants in hacknights list.

w1stler: @stanislawK still valid (using exclude)

backend/serializers/hacknight_serializer.py
```diff
@@ -0,0 +1,17 @@
...
+        many=True
+    )
+
+
+hacknight_schema = HacknightSchema()
+hacknights_schema = HacknightSchema(many=True)
```
w1stler: This two to be removed.

backend/serializers/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .participant_serializer import ParticipantSchema
```
magul: I understand, that we need to do that because `HacknightSchema` is not registered in Marshmallow registry, but think that we should either move all serializers to `backend/serializers.py` (as there not much of them and don't see a reason to have them in a package instead of a module) or at least us `__all__` and import all schemas from siblings modules.

backend/serializers/hacknight_serializer.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class HacknightSchema(Schema):
+    class Meta:
+        fields = ('id', 'date', 'participants')
+        dump_only = ('id',)
+
+    participants = fields.Nested(
+        'ParticipantSchema',
+        exclude=('name', 'lastname', 'email', 'hacknights', 'phone'),
```
magul: I thought that we want to exclude only hacknigths here (just to not create circular dependency).

backend/serializers/__init__.py
stanislawK: There was a discussion on this topic, and the decision was to keep serializers separately. So I sticked to that. We have hacknights nested in participants schema, and participants nested in hacknights schema. It caused circular imports, and that why I imported both schemas in __init__.py - to avoid circular import error.

backend/serializers/hacknight_serializer.py
stanislawK: I believe that You suggested me, that we shouldn't return whole participant object in hacknight schema. That why I used only with 'id' and 'github' arguments, which I changed then after Your suggestion to exclude.

PR #116, opened 2019-07-09
9 header component [WIP]

kaczynska:


frontend/src/components/Header.vue
```diff
@@ -0,0 +1,75 @@
...
+.active {
+  opacity: .8;
+}
+.v-btn {
+  font-weight: 700;
+}
+.v-menu {
+  margin-right: 20px;
+}
+</style>
```
w1stler: No newline

w1stler:
Do we want to have logo in header instead of plain text? Currently it's this way. We should also remove the white box beneath the header.

kaczynska:
> Do we want to have logo in header instead of plain text? Currently it's this way. We should also remove the white box beneath the header.
> 
> https://ibb.co/0265wWP
> https://ibb.co/RCxjcY1

We want to use the svg format because it is optimal and completely compatible with the original version of the character. When using text, we do not have control over the distance between letters, and these in practice are often adjusted manually (at the stage of working on the sign).

The white box should of course be removed.

frontend/src/assets/logo-white.svg
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
```
magul: if we're going to have logos in SVG can we also store them in https://github.com/CodeForPoznan/Community/tree/master/logos

w1stler:
Closing - @stanislawK is working on this in #227

PR #127, opened 2019-07-17
Travis-CI setup (fixes #212)

magul:


magul:
@adamgajzlerowicz removed `appveyor.yml` as Travis was chosen on last planning and it also looks that we AppVeyor was not yet configured (and Travis is already working).

PR #142, opened 2019-07-31
rename participant field (fixes #136)

### LINKED ISSUE #136: Rename participant field from name to first name and lastname to last name
In Python use `first_name`/`last_name`, in JavaScript use `firstName`/`lastName`.

kristhina:
I hope that it is correct. I don't know how to test it.


backend/tests/test_db.py
```diff
@@ -29,18 +29,18 @@ def test_add_new_participant_to_db(_db, new_participant):
     assert len(db.session.query(Participant).all()) == 0
 
     new_participant = Participant(
-        name=new_participant['name'],
-        lastname=new_participant['lastname'],
+        first_name=new_participant['first_name'],
```
magul: We could do that as `new_participant = Participant(**new_participant)`

backend/tests/test_models.py
```diff
@@ -19,14 +19,14 @@ def test_create_new_user(new_user):
 def test_create_new_participant(new_participant):
     """Test participant model."""
     new_participant = Participant(
-        name=new_participant['name'],
-        lastname=new_participant['lastname'],
+        first_name=new_participant['first_name'],
```
magul: Ditto: `new_participant = Participant(**new_participant)`

backend/migrations/versions/ea7204474ffa_.py
```diff
@@ -26,7 +26,7 @@ def upgrade():
     op.create_table('participant',
     sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
     sa.Column('name', sa.String(length=20), nullable=True),
-    sa.Column('lastname', sa.String(length=20), nullable=True),
+    sa.Column('last_name', sa.String(length=20), nullable=True),
```
magul: Historical cannot be changed.

backend/factories.py
```diff
@@ -25,13 +25,13 @@ class ParticipantFactory(BaseFactory):
     class Meta:
         model = Participant
 
-    name = factory.Faker("first_name", locale="pl_PL")
-    lastname = factory.Faker("last_name", locale="pl_PL")
+    first_name = factory.Faker("first_name", locale="pl_PL")
+    last_name = factory.Faker("last_name", locale="pl_PL")
     email = factory.LazyAttribute(
-        lambda obj: "{}@cfp.com".format(obj.lastname)
+        lambda obj: "{}@cfp.com".format(obj.last_name)
```
magul: We don't own `cfp.com` domain, it's not connected with us anyhow. We should change domain here to `codeforpoznan.pl` od stick to reserved, testing domains (like `example.com`, or `codeforpoznan.test`).

backend/migrations/versions/98e8c06e83ba_.py
```diff
@@ -0,0 +1,34 @@
...
+# revision identifiers, used by Alembic.
+revision = '98e8c06e83ba'
+down_revision = '7408c872d6c0'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('participant', sa.Column('first_name', sa.String(length=50), nullable=True))
```
magul: We should use `alter_column` here, not `drop_column` and `add_column` as altering will preserve data.

backend/factories.py
w1stler: .test is the best option

backend/tests/test_db.py
kristhina: changed :)

backend/tests/test_models.py
kristhina: changed :)

backend/migrations/versions/ea7204474ffa_.py
kristhina: changed to lastname

backend/factories.py
kristhina: changed to "codeforpoznan.test"

backend/migrations/versions/98e8c06e83ba_.py
kristhina: It has been done automatically by alembic

w1stler: Not sure about this one with alembic. 😕

kristhina: @magul, we are waiting for your re-review

magul: I understand, that Alembic could create such migration, but that's just a tool and in that case, it does it wrong. We're not on production right now, but as a learning exercise, we should prepare that migration in such manner, that it could be safely run on a production environment.

Right now that migration will drop existing columns an create new ones. Data previously stored in removed columns are permanently lost. I belive we should use here https://alembic.sqlalchemy.org/en/latest/ops.html?highlight=unique#alembic.operations.Operations.alter_column

kristhina: @magul 
I made all the requested changes in migration files. I checked the command flask db upgrade and flask db migrate - everything worked. When I opened the database in dbeaver it had column names: first_name and last_name. I hope that now everything works ok, but it would be nice if you could check :)

backend/migrations/versions/79060a1bab8e_.py
```diff
@@ -18,10 +18,7 @@
 
 def upgrade():
     # ### commands auto generated by Alembic - please adjust! ###
-    op.alter_column('hacknight', 'date',
-               existing_type=sa.VARCHAR(length=20),
-               type_=sa.Date(),
-               existing_nullable=True)
+    op.execute("""ALTER TABLE hacknight ALTER COLUMN date TYPE DATE USING date::date""")
```
magul: And what happened here?

kristhina: @magul there was some problem with the upgrade of database with date type. It required: "USING date::date" and in SQLAlchemy it was impossible to add this "USING" part. With @stanislawK we were looking for different solutions and we found that we should use pure SQL to solve this problem.

Issue #97, opened 2019-06-05
Find a CC0 license image of magnifier glass


OtisRed:
In the assets directory (introduced in #88) there is an image of magnifier glass that is rendered when you hover the mouse over the project image. 

This image comes from wikipedia and is licensed with CC BY 3.0 which means that in order to use it we have to credit the author in some visible place which is aesthetically troublesome for a landing page.

Therefore we need a similar image that is licensed for the public unlimited use (licences as CC0 or MIT), that we could deploy instead.

Steps to do:
1) Find an image of magnifying glass, which can be used based on CC0, MIT or similar license
2) Switch the images in the assets/images/ directory
3) Remember to change the image reference in ourProjects

Tomasz1234:
I will do it

magul:
Done in #245

PR #151, opened 2019-08-08
Modal content (fixes #113)

### LINKED ISSUE #113: Create modal content for our projects
Create a child component to our-projects-component that will render modal content after clicking particular project tile to view corresponding details as:
* Project title
* Project description

It shall also contain the "close" button.

OtisRed:
Complete although a lil bit messy in terms of css, so any suggestions (@Loczi94) in that regard would be great.

**Manual testing:**
1. docker up the app
1. click any of the projects
1. check if it renders proper details (use Volontulo - the rest doesn't have all the data filled yet)
1. check if all buttons work
1. check the console for errors


OtisRed:
@stanislawK @jacekkalbarczyk it's a workin prototype, although unstylized (I'm not really good with scss). You will find the most representative example under Volontulo project.

frontend/src/components/ModalContent.vue
```diff
@@ -0,0 +1,151 @@
...
+        onClick() {
+            this.$root.$emit('close');
+        }
+    }
+};
+</script>
+
+<style lang="scss" scoped>
+@import './../main.scss';
+
```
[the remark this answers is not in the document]
OtisRed: Thanks! didn't know that one

frontend/src/components/ModalContent.vue
```diff
@@ -0,0 +1,151 @@
...
+    </v-card>
+</template>
+
+<script>
+export default {
+    data() {
+        return {
+            copyleftIcon: require('@/assets/images/copyleft.svg')
+        };
+    },
```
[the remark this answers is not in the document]
OtisRed: Good catch!

frontend/src/components/ModalContent.vue
```diff
@@ -0,0 +1,151 @@
...
+    props: ['selectedProject', 'dialog'],
```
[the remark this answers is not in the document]
OtisRed: Good catch again

frontend/src/components/ModalContent.vue
```diff
@@ -0,0 +1,151 @@
...
+.content {
+    font-family: $font-content;
+    font-size: 1.5rem;
+    padding: 0rem 1rem 1rem 1rem;
+    line-height: 1.43;
+}
+
+.partner {
+    display: flex;
+    position: relative;
```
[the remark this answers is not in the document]
OtisRed: No i can't the titles within v-card-text falls out of line

OtisRed:
@Loczi94 the code after merging master and refactoring that came afterwards :D

frontend/src/components/OurProjects.vue
```diff
@@ -1,60 +1,201 @@
...
-        },
-        {
-          name: 'Polska Akcja Humanitarna',
-          image: ''
-        }
-      ]
+      dialog: false,
+      hoveredImg: require('@/assets/images/magnifier.svg'),
+      projects: projects_db,
+      // projects: [
```
magul: no commented code, please

frontend/src/components/OurProjects.vue
```diff
@@ -1,60 +1,201 @@
...
     </v-layout>
+    <v-dialog v-model="dialog" width="50rem">
+      <app-modal-content :selectedProject="selectedProject"></app-modal-content>
+    </v-dialog>
   </v-container>
 </template>
 
 <script>
+import ModalContent from './ModalContent.vue';
+import projects_db from '../assets/projects';
```
magul: I would say that changing the name of what is imported is so rare situation, that I would say that we should be fine with just `projects` here.

frontend/src/components/OurProjects.vue
```diff
@@ -78,7 +225,7 @@ export default {
   height: 100%;
 }
 
-#hoverd-img {
+#card--hover {
```
magul: Do we have a standard with two dashes in `id` name? Also, a fun fact is that we shall not introduce multiple ids on the same page (as that's not supported by browsers) and I see that you're using that id in `v-for` loop.

OtisRed: In this case yes. According to [BEM](http://getbem.com/naming/) two hyphens signify block modyfier as #card--hover modifies #card. The problem is rather that they should be a .class rather than #id as you've suggested.

frontend/src/components/OurProjects.vue
OtisRed: Forgot about it :(

frontend/src/components/OurProjects.vue
OtisRed: Yes it is. I wanted to differentiate both not to have var like projects: projects because I was afraid it won't work. It did so you were right.

OtisRed:
@magul if you check #97 there is already a task for that particular issue, although the rationale was different as current picture is shared on cc by 4.0 license instead of MIT or CC0. 

As I understand you are not a fan of putting things like this in static assets? 

Also I addressed the rest of the comments, thanks for them. If you're satisfied please approve not to block that PR :)

frontend/src/components/ModalContent.vue
```diff
@@ -0,0 +1,135 @@
...
+  text-transform: none;
+}
+
+.buttons {
+  display: flex;
+  align-content: center;
+  margin: 0.5rem;
+}
+
+#close-button {
```
Bosmanfrx: Do we want to use ids? They are quite dangerous, because they don't work when there are 2 same ids on the same page

frontend/src/components/ModalContent.vue
```diff
@@ -0,0 +1,135 @@
...
+  position: relative;
+  align-content: center;
+  font-family: $font-content;
+  font-size: 1.5rem;
+  line-height: 1.43;
+  font-weight: bold;
+}
+
+.stack-list {
+  background-color: $blue !important;
```
Bosmanfrx: !importan is evil :D

frontend/src/components/ModalContent.vue
```diff
@@ -0,0 +1,135 @@
...
+  color: $white;
+}
+
+#title-header {
```
Bosmanfrx: ditto

frontend/src/components/ModalContent.vue
```diff
@@ -0,0 +1,135 @@
...
+  background: $blue;
+  height: 7rem;
```
Bosmanfrx: do we want to set height of this element? Maybe content can decide about height?

frontend/src/components/OurProjects.vue
```diff
@@ -1,72 +1,91 @@
...
                 </div>
               </v-expand-transition>
             </v-card>
           </v-hover>
         </v-item-group>
-      </v-flex>
-    </v-layout>
+      </v-col>
+    </v-row>
+    <v-dialog v-model="dialog" width="50rem">
```
Bosmanfrx: that's huge width. Maybe percentage would be better?

frontend/src/components/ModalContent.vue
OtisRed: I KNOW! But I can't really get atound priorities in vuetify...

OtisRed: We've managed to get around this by finding exception from the situation when vuetify assert its own style as primary

frontend/src/components/ModalContent.vue
OtisRed: ditto-response :D

frontend/src/components/ModalContent.vue
OtisRed: I achieved the same thing by broadening the padding on the <p> element that contains title.

frontend/src/components/OurProjects.vue
OtisRed: It was there because of the large screen view when the dialog was getting extended throughout the screen making it look ridiculous. I changed it to max-width="50rem" to make it more elegant.

frontend/src/components/ModalContent.vue
Bosmanfrx: Do we really want to change the color though? Maybe we could customize theme of vuetify and use secondary button?

Bosmanfrx: Docs link: https://vuetifyjs.com/en/customization/theme

OtisRed: It's been transferred to #213

OtisRed:
@stanislawK from my part it's ready to merge - finally

PR #154, opened 2019-08-14
[107] constant width of components

Loczi94:
Vuetify classes were added to AboutUs component.
Global class container with max-width was created.

### DIFF
--- a/frontend/src/components/AboutUs.vue
+++ b/frontend/src/components/AboutUs.vue
@@ -13,7 +13,7 @@
                 xs12
                 md6
                 lg6
-                xl3>
+                xl4>
                     <v-card flat dark tile color="transparent">
                         <v-card-text>
                             KIM JESTEŚMY?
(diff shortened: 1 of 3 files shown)


OtisRed:
I've checked and for some reason deleting "fluid" attribute from v-container somehow doesn't box the blue background anymore. So it seems we should achieve the aim of this task by simply deleting "fulid" attribute from blue containers.

OtisRed:
You've got conflicts here, plus please make sure you merged current master into PR :)

Loczi94:
@OtisRed done :)

OtisRed:
@Loczi94 Now I've noticed that you've left fluid attribute in About us component :P Sorry, I didn't catch that before. I can delete it from here if you don't wanna look at it anymore :D

OtisRed:
As we upgraded vuetify it doesn't work anymore plus I have a strong feeling that it will be solved along all the scss mess :D

Issue #156, opened 2019-08-21
Unify docker container names


ngiersz:
Unify docker container names, so they have the same name on every machine

magul:
Done in #158

PR #158, opened 2019-08-21
[156] Container names set to cfp_v3_{container}_1

ngiersz:


w1stler:
Why are we changing that? ;-)

ngiersz:
Container names was sometimes created randomly (e.g. codeforpoznanpl_v3_backend_1 or code_for_poznan_pl_v3_backend_1 or codeforpoznanplv3_backend_1) and they were not coherent with these in Makefile

docker-compose.yml
```diff
@@ -3,6 +3,7 @@ version: "3"
 services:
 
   frontend:
+    container_name: cfp_v3_frontend_1
```
w1stler: We can remove `_1` from here as well as in other places.

Makefile
```diff
@@ -13,13 +13,13 @@ stop:	## Stop the environment
 	docker-compose stop
 
 bash:	## Go to the backend container
-	docker exec -ti codeforpoznanpl_v3_backend_1 bash
+	docker exec -ti cfp_v3_backend_1 bash
```
w1stler: Here we can use `docker-compose exec` with the namees `frontend`, `backend`, etc. Just please test it. ;-)

w1stler:
Fixes #156

PR #172, opened 2019-09-11
[62] Create API endpoint for hacknight details

w1stler:
To test, use Postman:
GET ```localhost:5000/hacknights/1/details/```


w1stler:
+ frontend test fixes for master branch

backend/app.py
```diff
@@ -40,3 +40,4 @@ def initialize_extensions(app):
 api.add_resource(ParticipantsList, "/participants/")
 api.add_resource(ParticipantDetails, "/participants/<int:id>/")
 api.add_resource(HacknightList, "/hacknights/")
+api.add_resource(HacknightDetails, "/hacknights/<int:id>/details/")
```
magul: also it shouldn't have `/details/` at the end

backend/resources/hacknight.py
```diff
@@ -44,3 +44,12 @@ def post(self):
 
         return {'message': 'Hacknight created successfully.',
                 "hacknight": data}, HTTPStatus.CREATED
+
+
+class HacknightDetails(Resource):
+    def get(self, id):
+        hacknight_schema = HacknightSchema(only=('date', 'id', 'participants'))
```
magul: why do we limit fields using `only` here?

w1stler: This was intentional, as the ticket stated this (only these fields to be returned). Currently we don't have others, so could be removed. :-)

magul: I believe it should be removed. I don't even see the word `only` in issue description so would assume that was to point out all fields in that model.

backend/tests/conftest.py
```diff
@@ -132,3 +132,12 @@ def add_hacknights(app, _db):
     for _ in range(10):
         HacknightFactory.create()
     _db.session.commit()
+
+
+@pytest.fixture
+def add_participants_to_hacknight(app, _db):
```
magul: We can simplify it:
```python
@pytest.fixture
def add_participants_to_hacknight(app, _db):
    HacknightFactory(participants=ParticipantFactory.create_batch(10))
    _db.session.commit()
```

PR #173, opened 2019-09-14
[83] Create refresh token endpoint

jacekkalbarczyk:
User login (auth/login/) endpoint is now returning both access and refresh token. Refresh token can be used to obtain new access token(via /auth/refresh/ route). There is a new endpoint (/auth/revoke-refresh-token/) to revoke refresh tokens (it should occur during logging out). I also added missing ending slashes to existing routes.
Use Postman app for tests.

Endpoints:

**Login**
  (method: POST) /auth/login/  
  takes raw JSON as a parameter:  
  {
      "username": username,
      "password": password
  }
If successful, returns access token, refresh token, HTTP status CREATED (201),
otherwise returns message "Not authorized", HTTP status UNAUTHORIZED (401).

**Refreshing access token**
(method: POST) /auth/refresh/  
  takes refresh token via Authorization header(Bearer token) as a parameter.  
  
If successful, returns new access token, HTTP status CREATED (201),
If refresh token was revoked returns message "token has been revoked", HTTP status UNAUTHORIZED (401).

**Revoking refresh token**
(method: DELETE) /auth/revoke-refresh-token/
  takes refresh token via Authorization header(Bearer token) as a parameter.  
  
If successful, returns message "Refresh token successfully revoked", HTTP status OK (200).

### DIFF
--- a/backend/resources/auth.py
+++ b/backend/resources/auth.py
@@ -98,8 +98,7 @@ class RefreshAccessToken(Resource):
             access_token, identity_claim
         )
 
-        ret = {'access_token': access_token}
-        return ret, HTTPStatus.CREATED
+        return {'access_token': access_token}, HTTPStatus.CREATED
 
 
 class RevokeRefreshToken(Resource):
--- a/backend/resources/auth.py
+++ b/backend/resources/auth.py
@@ -6,8 +6,9 @@ from backend.models import User, JWTToken
 from backend.serializers.login_serializer import LoginSchema
 from flask import current_app as app, request
 from flask_jwt_extended import (
-    create_access_token, decode_token,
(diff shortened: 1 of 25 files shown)


backend/resources/auth.py
```diff
@@ -66,6 +79,39 @@ def delete(self):
...
+
+        access_token = create_access_token(
+            identity=current_user, expires_delta=timedelta(minutes=60)
+        )
+        add_token_to_database(
+            access_token, identity_claim
+        )
+
+        ret = {'access_token': access_token}
+        return ret, HTTPStatus.CREATED
```
stanislawK: I believe we could just use here: return {'access_token': access_token}, HTTPStatus.CREATED

PR #174, opened 2019-09-15
Vuetify upgraded with all dependencies (fixes#171)

stanislawK:
To test: docker-compose build, and docker-compose up. Titles aren't style properly - elements with sass styling have to be fixed.


OtisRed:
@magul I think @stanislawK  responded to your comments

frontend/package.json
```diff
@@ -9,29 +9,32 @@
...
-    "vuetify": "1.5.14",
-    "vuex": "3.0.1"
+    "axios": "^0.18.1",
+    "sass": "^1.22.0",
+    "vue": "^2.6.10",
+    "vue-router": "^3.0.2",
+    "vuelidate": "^0.7.4",
+    "vuetify": "^2.0.0",
+    "vuex": "^3.0.1",
+    "webpack": "^4.39.3"
```
magul: how does it work before `webpack` in dependencies?

frontend/package.json
```diff
@@ -9,29 +9,32 @@
...
   },
   "dependencies": {
-    "axios": "0.18.1",
-    "vue": "2.6.10",
-    "vue-router": "3.0.2",
-    "vuelidate": "0.7.4",
-    "vuetify": "1.5.14",
-    "vuex": "3.0.1"
+    "axios": "^0.18.1",
+    "sass": "^1.22.0",
```
magul: as I understand it replaced `node-sass`? Why not in `devDep...`?

frontend/package.json
```diff
@@ -9,29 +9,32 @@
...
+    "vue": "^2.6.10",
+    "vue-router": "^3.0.2",
+    "vuelidate": "^0.7.4",
+    "vuetify": "^2.0.0",
+    "vuex": "^3.0.1",
+    "webpack": "^4.39.3"
   },
   "devDependencies": {
     "@vue/cli-plugin-babel": "3.8.0",
```
magul: no `^`?

frontend/package.json
OtisRed: Maybe here's the reason why scss stopped overriding vuetify classess...

magul: I doubt that as it wasn't yet merged.

OtisRed: And scss doesn't work only on this particular branch. It works fine on master

magul: I didn't anticipate, that you test branches before merging.

OtisRed: It seems like either compliment or insult simultaneously :P

frontend/package.json
stanislawK: We don't use vuetify-loader (webpack plugin) right now. I don't know if it wasn't required in older version of vuetify, or it works becaouse we have simple components structure. Nevertheless when i tried to add vuetify classes in dashboard subcomponent, vuetify couldn't load them correctly. In addition vuetify-loader is recomended solution for treeshaking in current vuetify docs https://vuetifyjs.com/en/customization/a-la-carte

frontend/package.json
stanislawK: vuetify from version 2.0 requires sass instead of node-sass, so i didn't see point to keep them both

frontend/src/main.js
```diff
@@ -1,11 +1,12 @@
...
-import Vuetify from 'vuetify';
+import Vuetify from 'vuetify/lib';
 import 'vuetify/dist/vuetify.min.css';
 
 import router from './router';
 import store from './store';
+import vuetify from './plugins/vuetify';
 
 Vue.use(Vuelidate);
 Vue.use(Vuetify);
```
magul: Isn't that a duplicate of what we have in `frontend/src/plugins/vuetify.js`?

PR #185, opened 2019-10-09
delete participant endpoint (fixes #170)

### LINKED ISSUE #170: create delete participant endpoint
Endpoint to delete participant from the database

kristhina:


backend/resources/participant.py
```diff
@@ -47,3 +47,11 @@ def get(self, id):
...
             Participant.query.get_or_404(id)
         ), HTTPStatus.OK
+
+    @jwt_required
+    def delete(self, id):
+        participant = Participant.query.get_or_404(id)
+        db.session.delete(participant)
+        db.session.commit()
+
+        return {"message": "Participant deleted successfully."}, HTTPStatus.OK
```
w1stler: No newline

kristhina: Newline added

backend/resources/participant.py
```diff
@@ -47,3 +47,12 @@ def get(self, id):
...
+
```
magul: There are some whitespaces in the last line (which should not be here).

PR #187, opened 2019-10-18
[74] implement black (fixes #68 #74)

w1stler:
`make rebuild`
`make test`

PR #184, opened 2019-10-09
[83] fixes after code review

jacekkalbarczyk:
Added NoResultFound error handling to sql queries.
Removed unused identity claim references.
Changed refresh token resource name and url. 
Added missing line (test_participant_details.py).

PR #191, opened 2019-10-23
Added tests related to refresh token (fixes #190)

### LINKED ISSUE #190: Tests for refresh token
Add tests for refresh token, and revoke refresh token

stanislawK:
make test


backend/tests/test_hacknight.py
```diff
@@ -2,23 +2,23 @@
...
 
 def test_get_hacknights_when_logged_in(
-    client, access_token, add_hacknights
+    client, tokens, add_hacknights
 ):
     """Test get list of hacknights for logged in user."""
     rv = client.get(
         '/hacknights/',
-        headers={'Authorization': 'Bearer {}'.format(access_token)}
+        headers={'Authorization': 'Bearer {}'.format(tokens["access"])}
```
Maciek246: If we're using Python 3.6+: Maybe f-string?

backend/tests/test_hacknight.py
```diff
@@ -2,23 +2,23 @@
...
     assert len(response['hacknights']) == 10
 
 
-def test_get_hacknights_with_empty_db(client, access_token):
+def test_get_hacknights_with_empty_db(client, tokens):
     """Test get list of hacknights when no hacknight added to db."""
     rv = client.get(
         '/hacknights/',
-        headers={'Authorization': 'Bearer {}'.format(access_token)}
+        headers={'Authorization': 'Bearer {}'.format(tokens["access"])}
```
Maciek246: If we're using Python 3.6+: Maybe f-string?
```suggestion
        headers={'Authorization': f'Bearer {tokens["access"]}'}
```

backend/tests/test_hacknight.py
stanislawK: I wanted to, but unfortunately we are on 3.5.2 right now. #187 will change python version to 3.8

magul:
We're waiting for #184 here but can you, please, fix tests that are failing and resolve conflicts here?

backend/tests/test_logout.py
```diff
@@ -2,19 +2,32 @@
...
                        headers={'Authorization': 'Bearer {}'
-                                .format(access_token)})
+                                .format(tokens["access"])})
     response = rv.get_json()
     assert rv.status_code == HTTPStatus.OK
     assert response['msg'] == 'Successfully logged out'
 
 
-def test_logout_user_without_token(client):
+@pytest.mark.parametrize('url', ['/auth/logout/', '/auth/revoke-refresh-token/'])
```
magul: That will change when #184 will get merged, but besides that, it's a pretty good change.

w1stler:
"Related", not "releted" (title). ;-)

stanislawK:
> It needs to be changed after #184.
> 
> Also (that's probably topic for different pull request) - what do you think about adding pytest fixture for authorized client? That way we could ommit almost everywhere this:
> 
> ```
> headers={'Authorization': 'Bearer {}'.format(tokens["access"])}
> ```

Good idea. I just introduced fixture with logged in client.

magul:
Unfortunately, I merged Black introduction by @w1stler and we have a lot of conflicts here. Can you merge `master` branch back here and ensure that Black is passing without reformating?

PR #194, opened 2019-10-30
Added tests for delete participant endpoint (fixes #193)

### LINKED ISSUE #193: Tests for delete participant endpoint
Write some tests for delete participant endpoint

stanislawK:
- make start
- make test


magul:
Can you, @stanislawK, merge back `master` here as #185 was merged and it would be great to see these 2 tests passing.

backend/tests/test_participant_details.py
```diff
@@ -1,36 +1,44 @@
...
 
-def test_get_non_existent_participant(
-    client, access_token, add_participants
+@pytest.mark.parametrize("method", ["get", "delete"])
+def test_get_delete_non_existent_participant(
+    client, access_token, add_participants, method
 ):
-    """Test get detais of non-existent participant"""
-    rv = client.get(
+    """Test get and delete non-existent participant"""
```
w1stler: Period at the end.

w1stler: Non-existing

backend/tests/test_participant_details.py
```diff
@@ -1,36 +1,44 @@
...
         headers={'Authorization': 'Bearer {}'.format(access_token)}
     )
     assert rv.status_code == HTTPStatus.NOT_FOUND
 
 
-def test_get_participant_unauthorized(client, add_participants):
-    """Test get participant details for not logged in user."""
+@pytest.mark.parametrize("method", ["get", "delete"])
+def test_get_delete_participant_unauthorized(client, add_participants, method):
+    """Test get and delete participant with not logged in user."""
```
w1stler: I would remove get from all the scenarios and ensure that we have explicit one for testing `get` method. Others will be eg.: "Test delete participant when user logged-in / logged-out".

stanislawK: I think it would make tests less readable, especially if we would merge #191. Following magul suggesion I introduced there logged in client fixture. So e.g. for testing get method I will have to inject two different test clients, which could be confussing.

OtisRed:
@magul or @w1stler please respond and/or approve because this one seems like it could be merged

PR #200, opened 2019-11-02
Additional tests for login endpoint (fixes #199)

### LINKED ISSUE #199: Add tests for login user
Add tests for situations where:
- User is  trying login twice
- Is no data in request body
- Data in request body is invalid
- One of values in request body missing

stanislawK:


backend/tests/test_login.py
```diff
@@ -36,3 +34,55 @@ def test_login_with_invalid_name_password(client):
...
+def test_try_login_twice(client, new_user, access_token):
+    """Test try login these same user twice."""
+    rv = client.post(
+        '/auth/login/',
+        json=new_user,
+        headers={'Authorization': 'Bearer {}'.format(access_token)}
+    )
+
+    response = rv.get_json()
+    assert rv.status_code == HTTPStatus.UNAUTHORIZED
```
magul: I'm going probably to merge this pull request as tests are passing but I would propose to consider if that's the best status code semantically speaking.

What do you think about `403 FORBIDDEN` as user is authorized to the system but forbidden to use this particular resource twice? We can also send more generic `400 BAD REQUEST` or even `200 OK` with proper authorization secrets as that doesn't look like a big deal here).

PR #202, opened 2019-11-03
Create new hacknight button (fixes #137)

### LINKED ISSUE #137: Create new hacknight button
Create button in dashboard, which will:
1) Create new hacknight (POST today's date and empty array of participants to datatbase)
2) Get that hacknight (GET this newly created hacknight by filtering today's date)
3) Redirect to add participants component

stanislawK:
At first populate database with command make populate_database.
After populating db there should be 30 hacknights dates in dropdown already.
Clicking add button should create new hacknight with todays date:
- select field in dropdown should be set to new hacknight
- new hacknight should be added to dropdown items
Clicking these same button second time, should result in show error alert, with the right message.


magul:
Can you resolve conflicts before I will jump into review here?

frontend/src/components/Hacknight.vue
```diff
@@ -0,0 +1,65 @@
...
+    </v-row>
+  </v-container>
+</template>
+
+<script>
+export default {
+  data() {
+    return {
+      selectedHacknight: '',
+      hacknight: {},
```
magul: shouldn't that be
```
selectedHacknight: undefined (or null),
hacknight: undefined (or null),
```
?

frontend/src/components/Hacknight.vue
```diff
@@ -0,0 +1,65 @@
...
+  },
+  created() {
+    this.$store.dispatch('hacknight/getHacknights').then(res => {
+      this.hacknights = res.data.hacknights;
+    });
+  },
+  methods: {
+    onCreateHacknight() {
+      this.$store.dispatch('hacknight/createHacknight').then(res => {
+        this.hacknight = res.data.hacknight;
```
magul: After digging into Vuex documentation I believe we should use its store in a different way than presented here.

Probably there's no reason to have `hacknights` as a component attribute, we could use the store directly here. It would probably also allow us to use the state without explicitly handling promises here.

magul: We can have that conversation on next hacknight @stanislawK

frontend/src/store/modules/hacknight.js
```diff
@@ -0,0 +1,67 @@
...
+            },
+            error => {
+              commit('raiseError');
+              reject(error);
+            }
+          );
+      });
+    },
+    getHacknight({ commit }, hacknight_id) {
+      return new Promise((resolve, reject) => {
```
magul: As `axios` claims to be
> Promise based HTTP client for the browser and node.js

there's probably no reason to encapsulate it inside `new Promise()` and you can simply write that down as
```js
return axios.get(`/hacknights/${id}/`).then((res) => {
  commit('wasSuccess');
  return res;
});
```
(error part is ommitted, as `axios` should raise that error by itself).

frontend/src/main.js
```diff
@@ -3,13 +3,21 @@ import App from './App.vue';
...
 
+import axios from 'axios';
 import router from './router';
 import store from './store';
 import vuetify from './plugins/vuetify';
 
 Vue.use(Vuelidate);
 
 Vue.config.productionTip = false;
+axios.defaults.baseURL = 'http://0.0.0.0:5000/';
```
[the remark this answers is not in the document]
stanislawK: As I said, I think that interceptors will be great to deal with refreshing tokens. I another cases, they could be quite confusing, for me at least. They changes app behevior in implicite way, and if they coused some bug, it would be extreme hard to track.
You could override axios defoult headers, quite expicitly, inside specific request, but I am not sure if it is some how possible with iterceptors.
I would be more then happy if more people could contribute, to that part of app, and I want to encourage them some how. I am affraid that adding more and more leyers of complexity will have exactly opposite effect.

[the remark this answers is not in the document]
OtisRed: That's true. I'm already getting lost at the level of vuex and axios so @stanislawK got a point on this one

OtisRed:
@stanislawK "how to test" description is kind of lousy here :P please describe briefly what would you expect the tester to do as the initial outline for testing as we agreed

backend/serializers/hacknight_serializer.py
```diff
@@ -9,3 +9,4 @@ class Meta:
     participants = fields.Nested(
         "ParticipantSchema", exclude=("hacknights",), many=True
     )
+    date = fields.Date()
```
OtisRed: What happens here?

stanislawK: Serialization from date js object to python date format

frontend/src/components/Hacknight.vue
```diff
@@ -0,0 +1,65 @@
...
+      this.$store
+        .dispatch('hacknight/getHacknight', this.selectedHacknight)
+        .then(res => {
+          this.hacknight = res.data;
+        });
+    }
+  }
+};
+</script>
+<style lang="scss" scoped>
```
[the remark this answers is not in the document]
OtisRed: Addressed in #235

frontend/src/components/Hacknight.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+  <v-container fluid text-xs-center>
+    <v-row align="center" justify="center">
```
OtisRed: To avoid in-line styling instead you could use:

`<v-row class="X`></row>`

And in style section

`<style lang="scss" scoped>`
`.X {`
`  display: flex;`
`  align-content: center;`
`  justify-content: center;`
`}`
`</style>`

frontend/src/components/Hacknight.vue
```diff
@@ -0,0 +1,70 @@
...
+      <v-col class="d-flex" cols="12" sm="6">
```
OtisRed: You can get the same result with:
`<v-col cols="6">`

frontend/src/components/Hacknight.vue
```diff
@@ -0,0 +1,70 @@
...
+        <v-alert
+          type="error"
+          :value="getError ? true : false"
+          transition="slide-y-transition"
+          dismissible
+          >{{ getError }}</v-alert
+        >
+      </v-col>
+    </v-row>
+    <v-row align="center" justify="center">
```
OtisRed: Ditto

frontend/src/components/Hacknight.vue
```diff
@@ -0,0 +1,70 @@
...
+      <v-col class="d-flex" cols="12" sm="6">
```
OtisRed: Ditto

frontend/src/components/Hacknight.vue
```diff
@@ -0,0 +1,70 @@
...
+        <v-select
+          v-model="selectedHacknight"
+          :items="getHacknights"
+          item-text="date"
+          item-value="id"
+          label="Select Hacknight"
+          @input="onGetHacknight"
+        >
+          <template v-slot:append-outer>
+            <v-btn
```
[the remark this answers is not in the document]
OtisRed: To avoid in-line styling, instead of current <v-btn/> i would use:

`<v-btn`
`  text`
`  class="Y"`
`  @click="onCreateHacknight"`
`>`

and in the <style> section:

`.Y {`
`  align-content: center;`
`  background-color: $blue;`
`  color: $white;`
`  display: flex;`
`  justify-content: center;`
`  margin: 4px;`
`  top: -12px;`
`}`

It assumes we get rif off things that seem do nothing :)

OtisRed:
@magul most of your comments got outdated along the corrections. If you have no further requests please unblock this PR.

frontend/src/components/Hacknight.vue
stanislawK: It is not inline styling - these are props provided by Vuetify.

frontend/src/main.js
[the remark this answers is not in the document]
OtisRed: Have we solved this?

frontend/src/components/Hacknight.vue
OtisRed: @stanislawK has that been resolved?

frontend/src/main.js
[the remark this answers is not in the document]
stanislawK: I think so - like I wrote, I don't feel super comfortable with interceptors, that why in that case axios defaults are safer.

frontend/src/components/Hacknight.vue
stanislawK: I hope so - I rewrote store implementation for Hacknight

PR #203, opened 2019-11-06
Fixing scss [WIP]

OtisRed:
@Loczi94

### DIFF
--- a/frontend/src/scss/global.scss
+++ b/frontend/src/scss/global.scss
@@ -4,19 +4,27 @@
 a {
     text-decoration: none;
 }
+// Container classes
 
-#title {
-    font-family: $font-header;
-    font-size: 42px;
-    position: relative;
-    padding-bottom: 3rem;
+#blue-container {
+    background: $blue;
+    padding: 1rem 2rem;
 }
+
+#white-container {
+    background: $white;
+    padding: 1rem 2rem;
+}
+
+// Container title classes
 
 #white-title {
     font-family: $font-header;
     font-size: 42px;
-    position: relative;
-    padding-bottom: 3rem;
+    color: $white;
+    display: flex;
+    justify-content: center;
+    padding: 3rem;
 }
 #white-title::before {
     content: '';
@@ -27,7 +35,7 @@ a {
     height: 5px;
     background-color: $white;
     margin-left: -125px;
-    margin-bottom: 3rem;
+    margin-bottom: 1rem;
 }
 
 #black-title {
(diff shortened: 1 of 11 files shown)


frontend/src/scss/global.scss
```diff
@@ -5,45 +5,72 @@ a {
...
 
-#white-title {
+.white-title,
+.blue-title {
     font-family: $font-header;
-    font-size: 42px;
-    position: relative;
-    padding-bottom: 3rem;
+    font-size: 2.8rem;
+    color: $white !important;
```
Loczi94: Using `important!` is not a good idea and there is a way to avoid it here.

frontend/src/scss/global.scss
```diff
@@ -5,45 +5,72 @@ a {
...
+    font-family: $font-header;
+    font-size: 1.2rem;
+    color: $white;
+    font-weight: bold;
+    display: flex;
+    justify-content: space-around;
+}
+
+.content {
+    color: $white !important;
```
Loczi94: ditto

frontend/src/scss/global.scss
```diff
@@ -5,45 +5,72 @@ a {
...
-    position: absolute;
-    bottom: 0;
-    left: 50%;
-    width: 250px;
-    height: 5px;
-    background-color: $white;
-    margin-left: -125px;
-    margin-bottom: 3rem;
+.blue-title {
+    color: $blue !important;
```
Loczi94: ditto

frontend/src/scss/global.scss
```diff
@@ -5,45 +5,72 @@ a {
...
-    font-family: $font-header;
-    font-size: 42px;
-    position: relative;
-    padding-bottom: 3rem;
+// Container classes
+
+.blue-container,
+.white-container {
+    background: $blue;
+    padding: 2rem 6rem;
```
Loczi94: I don't like the idea of defining these two classes together since they have only common padding. And you have to override color of `white-container`.

frontend/src/scss/global.scss
```diff
@@ -5,45 +5,72 @@ a {
...
+    padding: 3rem;
+
+    &::before {
+        content: '';
+        position: absolute;
+        bottom: 0;
+        left: 50%;
+        width: 250px;
+        height: 5px;
+        background-color: $white;
```
Loczi94: ditto - You define `background-color` here and then you have to override it for `blue-title` which doesn't make any sense.

Loczi94:
Done in issue 164 (PR: 211).

PR #212, opened 2019-11-13
create edit participant endpoint (fixes #61)

### LINKED ISSUE #61: Create edit participant endpoint

kristhina:
Testing:
- make start
- make populate_database
- login in postman with given username: "/auth/login"
- copy access_token
send "put" request to "participants/1/":
Headers: Authorization: Bearer access_token; Content-Type: application/json
Body:
{
	"first_name" : "janusz007",
	"github": "januszek2"
}


kristhina:
Status code changed to 200
https://stackoverflow.com/questions/2342579/http-status-code-for-update-and-delete

backend/resources/participant.py
```diff
@@ -55,4 +57,19 @@ def delete(self, id):
...
+        try:
+            data = participant_schema.load(json_data)
+        except ValidationError as err:
+            return (err.messages), HTTPStatus.BAD_REQUEST
+
+        for key, value in data.items():
+            setattr(participant, key, value)
+        db.session.add(participant)
+        db.session.commit()
+        return {"participant": participant_schema.dump(participant)}, HTTPStatus.OK
```
w1stler: Please remove: `"participant":`

backend/resources/participant.py
```diff
@@ -26,7 +28,7 @@ def post(self):
         json_data = request.get_json(force=True)
         if not json_data:
             return {'message': 'No input data provided'}, \
-                HTTPStatus.BAD_REQUEST
+                   HTTPStatus.BAD_REQUEST
```
w1stler: Won't that fit in one line?

backend/resources/participant.py
kristhina: done

backend/resources/participant.py
kristhina: changed to one line

backend/resources/participant.py
```diff
@@ -55,4 +56,19 @@ def delete(self, id):
...
+        participant_schema = ParticipantSchema()
+        participant = Participant.query.get_or_404(id)
+        json_data = request.get_json(force=True)
+        try:
+            data = participant_schema.load(json_data)
+        except ValidationError as err:
+            return (err.messages), HTTPStatus.BAD_REQUEST
+
+        for key, value in data.items():
+            setattr(participant, key, value)
```
magul: Also from Marshmallow docs - they use `make_object` with `@post_load` decorator to create DB object just by `participant = participant_schema.load(json_data)`. What do you think about that?

see: https://marshmallow.readthedocs.io/en/stable/examples.html#todo-api-flask-peewee
or here: https://github.com/marshmallow-code/marshmallow-sqlalchemy/blob/dev/src/marshmallow_sqlalchemy/schema.py#L183

magul: BTW is there any reason that we settled with `marshmallow` and not `marshmallow-sqlalchemy`? it looks like it could help us here a lot.

Here are a few questions that should be somehow addressed:
* should that endpoint update all fields in participant model? (meaning that if there's no key in json payload, it shoulld set that value to null?
* I understand the necessity of having the first name required, but if we will settle with partial updates (meaning that existing instance will be a base that we will update on) so do we really need to provide this particular field every time we want to update this participant (even if the first name will not change)?

OtisRed:
@magul please refer to the changes that has been made

PR #224, opened 2019-12-06
Adjusting to mobile view (fixes #223) 

### LINKED ISSUE #223: Adjust styling to mobile view
At the moment the website in mobile view is a complete mess, which includes problems with container paddings or breaking the title in two lines which overlap when are broken by the small screen.

OtisRed:
Introduces media query in scss that renders classes conditionally depending on the kind of screen that is being used. 

**How to test it**
1. Check for the errors in console
1. Test visually preferably using multiple devices view (might wanna use emmet re:view plugin)

### DIFF
--- a/frontend/src/scss/breakpoints.scss
+++ b/frontend/src/scss/breakpoints.scss
@@ -1,0 +1,1 @@
+$small-device: 600px;
--- a/frontend/src/scss/breakpoints.scss
+++ b/frontend/src/scss/breakpoints.scss
@@ -1,1 +1,2 @@
-$small-device: 600px;
+$phone: 600px;
+$desktop: 1264px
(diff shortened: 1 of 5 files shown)


frontend/src/scss/breakpoints.scss
```diff
@@ -0,0 +1,2 @@
+$phone: 600px;
```
Bosmanfrx: where did You get these breakpoints from?

OtisRed: I modeled them after: https://vuetifyjs.com/en/customization/breakpoints 
Do you @Bosmanfrx have better suggestion how to define breakpoints?

Bosmanfrx: I was just curious. ;) There is no one set of breakpoints. Maybe we could create map of all breakpoints here?

PR #229, opened 2019-12-29
Added remove_expired_tokens method (fixes #228)

### LINKED ISSUE #228: Remove expired tokens
There is no need to store in db expired tokens since that info is coded in token itself. That why it could be beneficial to have some method to clean up our db from expired tokens

stanislawK:
Testing:
Check if You have some expired tokens in db:
- make start 
- make psql
- inside psql: SELECT expires FROM jwt_tokens;
If you don't have there expired tokens you could navigate to /backend/resources/auth.py and in UserLogin change:

`            access_token = create_access_token(
                identity=username, expires_delta=timedelta(minutes=60)
            )`

to:
`            access_token = create_access_token(
                identity=username, expires_delta=timedelta(microseconds=1)
            )`

Then login via postman, or frontend form. That way you should be able to add token which expires after 1 microsecond.
Execute command: make remove-tokens.
Look for expired tokens in db via psql like in first steps. All tokens should have expires date in future - no expired token should left in database.


w1stler:
@stanislawK please resolve conflicts.

backend/models.py
```diff
@@ -80,3 +82,14 @@ def is_jti_blacklisted(cls, jti):
...
+
+    def expired(self):
+        return datetime.now() > self.expires
+
+    @classmethod
+    def remove_expired(cls):
+        tokens = cls.query.all()
+        for token in tokens:
+            if token.expired():
+                db.session.delete(token)
```
w1stler: It could be simplified:

```
for token in cls.query.all():
```

PR #230, opened 2019-12-31
Add participant to hacknight component v2 (fixes #55)

### LINKED ISSUE #55: Create component to Add participants to hacknight
Multi-select dropdown widget with save button (disable Enter key on save button).

slim-select

stanislawK:
At first populate database with command make populate_database.
After populating db there should be 30 hacknights dates in dropdown already.
Select one of those. List with that hacknight's participants should be displayed.
Try to add new participants from dropdown. After that they should appear on a list. You should not be able to select from dropdown participants which are on a list already.
Clicking add hacknight button should create new hacknight with todays date and empty array of participants.


backend/resources/hacknight.py
```diff
@@ -54,3 +48,35 @@ def get(self, id):
...
+
+        json_data = request.get_json(force=True)
+        ids_schema = Schema.from_dict({"participants_ids": fields.List(fields.Int())})
+        try:
+            data = ids_schema().load(json_data)
+        except ValidationError as err:
+            return err.messages, HTTPStatus.UNPROCESSABLE_ENTITY
+
+        new_participants = [
+            id for id in data["participants_ids"] if id not in participants
```
Maciek246: Probably nitpick but maybe could you rename `id` on `_id` or something similar to avoid overwrite python id function.

frontend/package.json
```diff
@@ -9,7 +9,9 @@
     "lint:fix": "vue-cli-service lint "
   },
   "dependencies": {
+    "@mdi/font": "^4.5.95",
     "axios": "^0.18.1",
+    "moment": "^2.24.0",
```
Bosmanfrx: Please don't go this way :(
We probably can get away just with native Date object, if not - date-fns is decent alternative
More info here ;)
https://github.com/you-dont-need/You-Dont-Need-Momentjs

frontend/src/components/Hacknight.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+  <v-container fluid text-xs-center>
+    <v-row align="center" justify="center">
+      <v-col class="d-flex" cols="12" sm="6">
+        <v-alert
+          type="error"
+          :value="getError ? true : false"
```
Bosmanfrx:
```suggestion
          :value="getError"
```

Bosmanfrx: Would it work?

frontend/src/components/Hacknight.vue
```diff
@@ -0,0 +1,70 @@
...
+          item-value="id"
+          label="Select Hacknight"
+          @input="onGetHacknight"
+        >
+          <template v-slot:append-outer>
+            <v-btn
+              color="blue-grey"
+              class="ma-2 white--text"
+              @click="onCreateHacknight"
+              style="top: -12px"
```
Bosmanfrx: should we add style like here in template? It mixes concerns a bit. Maybe we could move it to css?

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,111 @@
...
+          clearable
+          deletable-chips
+          append-outer-icon="mdi-cloud-upload"
+        >
+          <template v-if="noData" v-slot:no-data>
+            <v-list-item>
+              <v-list-item-content>
+                <v-list-item-title>
+                  No results matching "<strong>{{ search }}</strong
+                  >".
```
Bosmanfrx: That's very strange line break

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,111 @@
...
+                </v-list-item-title>
+              </v-list-item-content>
+            </v-list-item>
+          </template>
+          <template v-slot:append-outer>
+            <v-btn
+              color="blue-grey"
+              class="ma-2 white--text"
+              @click="onAddParticipants"
+              style="top: -12px"
```
Bosmanfrx: ditto

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,111 @@
...
+    };
+  },
+  created() {
+    this.$store.dispatch('participant/getParticipants');
+  },
+  methods: {
+    onAddParticipants() {
+      const ids = [];
+
+      for (const participant of this.selectedParticipants) {
```
Bosmanfrx: `const ids = this.selectedParticipants.map(participant => participant.id)`

frontend/src/components/HacknightsParticipants.vue
stanislawK: When i've changed it, prettier reformatted it as above again :/

frontend/src/components/HacknightsParticipants.vue
stanislawK: Neat - thanks

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,110 @@
...
+              participant =>
+                !getHacknight.participants.filter(y => y.id === participant.id)
+                  .length
+            )
+          "
+          item-text="email"
+          item-value="email"
+          :search-input.sync="search"
+          hide-selected
+          label="Add participants"
```
OtisRed: Inside the app we use Polish :P please change to "dodaj użytkowników"

The same goes for hacknight component line 31 where instead of "New" should stand "Nowy hacknight" - I didn't catch it then before merging.

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,110 @@
...
+          small-chips
+          clearable
+          deletable-chips
+          append-outer-icon="mdi-cloud-upload"
+        >
+          <template v-if="noData" v-slot:no-data>
+            <v-list-item>
+              <v-list-item-content>
+                <v-list-item-title>
+                  No results matching "<strong>{{ search }}</strong
```
OtisRed: Replace with polish counterpart

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,110 @@
...
+      </v-col>
+    </v-row>
+    <v-card class="mx-auto" color="white" max-width="500">
+      <v-toolbar flat color="transparent">
+        <v-avatar color="blue" size="48" style="margin-right:1em;">
+          <span class="white--text headline">{{
+            getHacknight.participants.length
+          }}</span>
+        </v-avatar>
+        <v-toolbar-title>Participants</v-toolbar-title>
```
OtisRed: Replace with polish counterpart

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,110 @@
...
+      </v-toolbar>
+      <v-divider v-if="getHacknight.participants"></v-divider>
+
+      <v-list>
+        <template v-for="(item, i) in getHacknight.participants">
+          <v-list-item v-if="getHacknight.participants" :key="i">
+            <v-list-item-avatar>
+              <v-icon>mdi-account-outline</v-icon>
+            </v-list-item-avatar>
+            <v-list-item-title v-text="item.email"></v-list-item-title>
```
OtisRed: Again, are we sure we want to use emails to entitle participants?

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,110 @@
...
+          </v-list-item>
+        </template>
+      </v-list>
+      <v-divider></v-divider>
```
OtisRed: Does this divider do anything? The list seems to be enclosed by the border of v-card

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,110 @@
...
+  </v-container>
+</template>
+
+<script>
+import { mapGetters } from 'vuex';
+export default {
+  data() {
+    return {
+      selectedParticipants: [],
+      search: null,
```
OtisRed: Maybe we could use some placeholder here? Like "wpisz imię uczestnika" or sth

frontend/src/store/modules/participant.js
```diff
@@ -0,0 +1,37 @@
...
+    getParticipants(state) {
+      return state.allParticipants;
+    },
+    getError(state) {
+      return state.error;
+    }
+  },
+  mutations: {
+    raiseError(state, error_msg) {
+      state.error = error_msg;
```
OtisRed: This is probably unnecessary but defaults error messages are rendered in polish. Although I don't think it's necessary to append them with our own messages.

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,110 @@
...
+              </v-list-item-content>
+            </v-list-item>
+          </template>
+          <template v-slot:append-outer>
+            <v-btn
+              color="blue-grey"
+              class="ma-2 white--text add-hacknight-btn"
+              @click="onAddParticipants"
+              offset-y
+            >
```
OtisRed: When you try to send empty list 400 error occurs. Maybe we should validate here?

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,110 @@
...
+          outlined
+          :items="
+            getParticipants.filter(
+              participant =>
+                !getHacknight.participants.filter(y => y.id === participant.id)
+                  .length
+            )
+          "
+          item-text="email"
+          item-value="email"
```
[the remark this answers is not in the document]
stanislawK: Which field then it should be? I discussed it with someone long time ago, and conclusion back then was that email is probably only field which we have for all participants.

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -0,0 +1,110 @@
...
+  <v-container fluid text-xs-center>
+    <v-row align="center" justify="center">
+      <v-col class="d-flex" cols="12" sm="6">
+        <v-combobox
+          v-model="selectedParticipants"
+          outlined
+          :items="
+            getParticipants.filter(
+              participant =>
+                !getHacknight.participants.filter(y => y.id === participant.id)
```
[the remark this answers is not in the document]
stanislawK: correct

frontend/src/components/HacknightsParticipants.vue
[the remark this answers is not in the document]
magul: I believe that the most common field that we have for participant is github login.

frontend/src/components/HacknightsParticipants.vue
stanislawK: Should we ? It will be inconsistent at least to the moment when we introduce some tool for internationalization to backend eg. Flask-Babel. But even now with english only we struggle to provide consistent messages to the frontent client. We could also intercept those messages and translate them somewhere within frontend, but until now no one came with such idea.

frontend/src/components/HacknightsParticipants.vue
stanislawK: I don't think it works as a placeholder: v-combobox label prop does.

frontend/src/components/HacknightsParticipants.vue
stanislawK: I added disabled prop based on whenever list of participants is empty or not

frontend/src/components/HacknightsParticipants.vue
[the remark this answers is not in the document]
OtisRed: Honestly I think I would never be able to make a complete list of attendence without bothering each individual person if I were to use either email or github login. Can't we just use names and surnames here? These most of us (whoever is going to make a list at a given time) know at least some of each other's names.

frontend/src/components/HacknightsParticipants.vue
OtisRed: I don't think I understand what you mean. Are those labels somehow related to messages? In the comment I assumed that they only display the string in the button and I figured that since (apart from github) we speak mostly in Polish it will be more natural. Also the potential inconsistency between languages in content and messages personally doesn't bother me at all, but maybe it's just me :)

frontend/src/components/HacknightsParticipants.vue
OtisRed: Sorry, I read the prop description wrong.

frontend/src/components/HacknightsParticipants.vue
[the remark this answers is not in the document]
OtisRed: Github login was suggested

PR #232, opened 2019-12-31
Added navigation guard for Dashboard component (fixes #231)

### LINKED ISSUE #231: Navigation Guard for Dashboard component
Navigation to Dashboard component should be possible only for logged in users. Add navigation guard in order to protect that route.

stanislawK:
Try go to http://localhost:8080/dasboard. You should be redirected to /login path.
Populate database with command make populate_database.

### DIFF
--- a/frontend/src/router.js
+++ b/frontend/src/router.js
@@ -4,13 +4,21 @@ import VueRouter from 'vue-router';
 import HomePage from './components/HomePage.vue';
 import Login from './components/Login.vue';
 import Dashboard from './components/Dashboard.vue';
+import store from './store/index';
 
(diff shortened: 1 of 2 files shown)


frontend/src/router.js
```diff
@@ -4,13 +4,21 @@ import VueRouter from 'vue-router';
 import HomePage from './components/HomePage.vue';
 import Login from './components/Login.vue';
 import Dashboard from './components/Dashboard.vue';
+import store from './store/index';
```
Maciek246: If this is only one 'index.js' file in this directory, JS should look for by default:
```suggestion
import store from './store';
```

Issue #213, opened 2019-11-13
Create vuetify theme for buttons


OtisRed:
Look n the comment - the problem is that we want to have our own blue to color different buttons. Right now we are doing it in the sort of hackish way in contact us component and modal content component - it would be great to standarzie this issue.

w1stler:
Sugmit? :>

OtisRed:
@OtisRed evaluate this

OtisRed:
Instead of moving this id into global class we could create theme according to vuetify documentation: https://vuetifyjs.com/en/customization/theme

MargoWM:
closed for now
@Loczi94 we gave up the idea

PR #234, opened 2020-01-13
Login, logout, refresh token (fixes #84)

### LINKED ISSUE #84: Login/logout with refresh token

stanislawK:
Proper testing will be possible after #202 or #230 merge.
Go to backend/resources/auth.py and in UserLogin and RefreshAccessToken change timedelta in:

`
            access_token = create_access_token(
                identity=username, expires_delta=timedelta(minutes=60)
            )
`

to something shorter eg. seconds=5. Then open network tab in devtools, login on /login, and navigate to /dashboard, your access_token should expired, and when you will try to access some protected resource eg. get hacknight, or participants, frontend should send request for new access token, and set it in headers automatically (check in network tab).
Logout is possible even with expired token.
When you in addition you will change refresh token expires_delta to shorter:

`
            refresh_token = create_refresh_token(
                identity=username, expires_delta=timedelta(weeks=1)
            )
`
after logging in and trying to access some protected resource again, both: access and refresh tokens should be removed from local storage, and you should not have access to protected resources, or to /dashboard even.


frontend/src/components/Login.vue
```diff
@@ -5,10 +5,10 @@
         <v-flex xs12 sm6>
           <v-alert
             type="error"
-            :value="error_msg ? true : false"
-            @click="error_msg = ''"
+            :value="getError ? true : false"
```
Maciek246: I assume you want to convert `getError` variable to a boolean value, so maybe you could use:
```suggestion
            :value="!!getError"
```

frontend/src/components/DashboardHeader.vue
```diff
@@ -0,0 +1,62 @@
...
+  }
+};
+</script>
+
+<style lang="scss" scoped>
+@import './../main.scss';
+.navbar-custom {
+  font-family: $font-header;
+  background-color: $blue-media-hr !important;
+  color: #ffffff !important;
```
Maciek246: Is these `!important`s required? Generally, this is "antipattern" and I believe it should be used only for overriding 3rd pard CSS. Is this a specific case?

Maciek246: Only proposition: maybe move all colors to separate file with variables and import?

frontend/src/components/DashboardHeader.vue
```diff
@@ -0,0 +1,62 @@
...
+    return {
+      image: require('@/assets/images/logo-white.svg')
+    };
+  },
+  methods: {
+    onLogout() {
+      this.$store.dispatch('auth/logout').then(() => {
+        this.$router.push('/');
+      });
+    }
```
Bosmanfrx:
```suggestion
    async onLogout() {
        await  this.$store.dispatch('auth/logout');
        this.$router.push('/');
    }
```

Bosmanfrx: That's just a suggestion. I'm not quite sure it'll work with Vue(it should though). By taking advantage of new async/await syntax (Present also in Python) we can simplify this a little. What do you think? ;)

frontend/src/router.js
```diff
@@ -4,13 +4,21 @@ import VueRouter from 'vue-router';
...
 
 const routes = [
   { path: '/', component: HomePage },
   { path: '/login', component: Login },
-  { path: '/dashboard', component: Dashboard }
+  {
+    path: '/dashboard',
+    component: Dashboard,
+    beforeEnter: (to, from, next) => {
+      if (!store.getters['auth/isLoggedIn']) next('/login');
```
Bosmanfrx: It's always good idea to add braces for `if` and `else`. We can always end as Apple: https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/ :P

frontend/src/store/modules/auth.js
```diff
@@ -1,42 +1,97 @@
 import axios from 'axios';
 
+import { interceptorRefresh } from '../../main.js';
+
 export default {
   namespaced: true,
   state: {
-    token: localStorage.getItem('token') || ''
+    token: localStorage.getItem('token') || '',
```
Bosmanfrx: What are requirements about browsers? I don't know much about Vue build process, but maybe (just maybe) we could use that shiny `localStorage.getItem('token') ?? ''` null coalescing syntax

frontend/src/store/modules/auth.js
```diff
@@ -1,42 +1,97 @@
...
+          localStorage.removeItem('refresh_token');
+        })
+        .catch(() => {
+          localStorage.removeItem('refresh_token');
+        });
+    },
+    refresh() {
+      const refresh_token = localStorage.getItem('refresh_token');
+
+      return axios
```
Bosmanfrx: Just general comment, I think this file could use async/await. It's up to you if it's more readable to you

frontend/src/components/DashboardHeader.vue
stanislawK: IMO it is very elegant solution. Thanks.

frontend/src/store/modules/auth.js
stanislawK: From docs it looks superb, but as You mentioned browser support is quite poorly, so I am not sure. @OtisRed what do You think ?

frontend/src/store/modules/auth.js
stanislawK: Should it looks like that?
```
      async login({ commit }, loginData) {
      delete axios.defaults.headers.common['Authorization'];
      try {
        let res = await axios
        .post('auth/login/', {
          username: loginData.username,
          password: loginData.password
        })
        const token = res.data.access_token;
        const refresh_token = res.data.refresh_token;
        localStorage.setItem('token', token);
        localStorage.setItem('refresh_token', refresh_token);
        axios.defaults.headers.common['Authorization'] = 'Bearer ' + token;
        return res.status;
      } catch(error) {
          localStorage.removeItem('token');
          commit('raiseError', error);
        };
    },
```

OtisRed:
Jest two pre-review remarks:
1) At this moment frontend validation requires 5 characters. I've seen minLength import in this PR but I didn't find where it is defined. Just a remminder - we've agreed to change minimum length validation to 3 characters.
2) At this moment there pressing "enter" isn't hooked to log in button. If this PR doesn't cover it, then it probably should.

frontend/src/components/DashboardHeader.vue
OtisRed: Check the comment in line 18.
If you replicate the solution from header component this will not be an issue.

frontend/src/components/DashboardHeader.vue
```diff
@@ -0,0 +1,62 @@
...
+    <v-app-bar class="navbar-custom" dark clipped-right>
+      <v-toolbar-title>
+        <a href="https://codeforpoznan.pl/">
+          <v-img
+            :src="image"
+            contain
+            max-height="20"
+            aspect-ratio="1.7"
+            max-width="500"
+            min-width="350"
```
OtisRed, line 12: Some older phones have 320px displays. E.g. I checked iPhone 4 and it fits all right but doesn't leave the space for hamburger icon if we decide to use one. I suggest set the min-width to 300px

frontend/src/components/DashboardHeader.vue
```diff
@@ -0,0 +1,62 @@
...
+            position="left"
+          ></v-img>
+        </a>
+      </v-toolbar-title>
+      <v-spacer></v-spacer>
+      <v-tabs
```
OtisRed, line 18: Please replicate the solution from header component

frontend/src/components/DashboardHeader.vue
```diff
@@ -0,0 +1,62 @@
...
+        background-color="transparent"
+        right
+        class="hidden-sm-and-down"
+        mobile-break-point="768px"
+      >
+        <v-tab align-right @click="onLogout">Logout</v-tab>
+      </v-tabs>
+    </v-app-bar>
+    <v-list class="navbar-custom hidden-md-and-up" dark>
```
OtisRed, line 27: Ditto - comment in line 18

frontend/src/components/DashboardHeader.vue
OtisRed: Probably themes could solve problems with overwriting
https://vuetifyjs.com/en/customization/theme/ 
We have a ticket #213 to pick up on that more systemically

frontend/src/components/Login.vue
```diff
@@ -79,29 +79,15 @@ export default {
...
-            } else if (err.includes('User already logged in')) {
-              this.error_msg = 'Jesteś już zalogowany';
-            } else if (error) {
-              this.error_msg = 'Logowanie nie powiodło się';
-            }
+        this.$store.dispatch('auth/login', loginData).then(status => {
+          this.showSpinner = false;
+          this.clearForm();
+          if (status == 201) {
+            this.successAlert = true;
```
OtisRed, line 88: You should push $router to /dashboard here, because typing URL gets really annoying :D

Also while you at it, maybe you could figure out how to trigger submitting form on enter. I tried @keyup.enter="onSubmit" and it didn't work for some reason.

frontend/src/components/Login.vue
```diff
@@ -54,14 +54,14 @@
...
 <script>
 import { minLength, required } from 'vuelidate/lib/validators';
+import { mapGetters } from 'vuex';
 
 export default {
   data() {
     return {
       username: '',
       password: '',
       showPass: false,
```
OtisRed, line 64: Can we rename it to showPassword? I needed to look for it to understand what it does.

frontend/src/store/modules/auth.js
```diff
@@ -1,42 +1,97 @@
...
+        .catch(() => {
+          localStorage.removeItem('token');
+        })
+        .finally(() => {
+          delete axios.defaults.headers.common['Authorization'];
+          dispatch('revoke');
+        });
+    },
+    revoke() {
+      const refresh_token = localStorage.getItem('refresh_token');
```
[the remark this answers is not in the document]
stanislawK: Access token is added as a default header in main.js, because we have to provide it in every request to endpoint required authentication. Refresh token is required only to refresh access token and to revoke itself. Since header for both access and refresh token is these same:
"Authorization": "Bearer {access_token}"
"Authorization": "Bearer {refresh_token}"
setting it with refresh_token doesn't make much sense. What is more important, we should consider different strategy to store refresh_token (I mentioned it on few hacknights). Refresh token lives long, so it is important to store it and use it safely. Here is one of possible solutions https://flask-jwt-extended.readthedocs.io/en/stable/tokens_in_cookies/

frontend/src/components/Login.vue
```diff
@@ -54,14 +54,14 @@
...
-      error_msg: '',
       successAlert: false,
       showSpinner: false
```
OtisRed, line 66: Could we rename for "showLoadingSpinner" or sth to make it more transparent?
Also spinner could be centered as right now it's aligned to the left

frontend/src/components/DashboardHeader.vue
```diff
@@ -0,0 +1,62 @@
...
+  <div>
+    <v-app-bar class="navbar-custom" dark clipped-right>
+      <v-toolbar-title>
+        <a href="https://codeforpoznan.pl/">
+          <v-img
+            :src="image"
+            contain
+            max-height="20"
+            aspect-ratio="1.7"
+            max-width="500"
```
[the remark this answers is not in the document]
stanislawK: When I removed "contain" logo looks like that:
Screenshot from 2020-04-22 20-29-51

frontend/src/components/Login.vue
stanislawK: Submitting form after enter press works, but redirect to Dashboard is surprisingly challenging, so I created new [ticket](https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/272) for that task

frontend/src/components/Login.vue
stanislawK: Done

frontend/src/components/DashboardHeader.vue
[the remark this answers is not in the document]
OtisRed: I meant you can delete aspect-ratio and max-width because contain does all the work and those are redundant. Contain should definitely stay.

frontend/src/components/Login.vue
OtisRed: Ok, so I close it

frontend/src/store/modules/auth.js
OtisRed: It's over my expertise to say sth meaningful here

frontend/src/components/DashboardHeader.vue
stanislawK: @OtisRed Do You thing it is still valid? I wasn't able to reproduce that. Hamburger menu is in place and working on 300px screen, changing min-width to suggested 300px, have no impact. Apart from that it is related to header, not to login/logout with refresh logic, which is these pr feature.

PR #238, opened 2020-01-31
Change string formatting (fixes #204) 

### LINKED ISSUE #204: Change string formatting
After python upgrade (pr #187) we could change string formatting from str.format() to f-strings.

halsing:
I changed string formatting from .format to f string in : 
* contact.py
* auth.py
* factories.py
* database settings in config.py

All tests passed, site is still working and I ran black to fix my code styling. 
Someone should check if messages in emails are still correct.


OtisRed:
As a good practice we have a convention to describe shortly how would you test it step by step. Not in a lengthy detail, but assuming that e.g. frontend guy like me could test it without much prior knowledge. 

Well, at least at the level of whether it works ok or not.

backend/config.py
```diff
@@ -17,12 +17,14 @@ class DevelopmentConfig(Config):
...
-    SQLALCHEMY_DATABASE_URI = "postgresql://{username}:{password}@{hostname}:{port}/{databasename}".format(
-        username=os.environ.get("DB_USER"),
-        password=os.environ.get("DB_PASSWORD"),
-        hostname=os.environ.get("DB_HOST"),
-        port=5432,
-        databasename=os.environ.get("DB_NAME"),
+    DB_USER = (os.environ.get("DB_USER"),)
+    DB_PASSWORD = (os.environ.get("DB_PASSWORD"),)
+    DB_HOSTNAME = (os.environ.get("DB_HOST"),)
+    DB_PORT = (5432,)
```
arturtamborski: I don't know much about this project and how it interacts with flask but I'm pretty sure that these variables shouldn't be tuples which they are now. The trick lies in placing comma right after the end of a value.
I'm almost certain that it's not what you want. Consider this example:
```
>>> x = 1,
>>> type(x)
<class 'tuple'>
```

backend/config.py
```diff
@@ -17,12 +17,14 @@ class DevelopmentConfig(Config):
...
+    DB_NAME = os.environ.get("DB_NAME")
+
+    SQLALCHEMY_DATABASE_URI = (
```
arturtamborski: Why is it wrapped around with braces `(<value>)` ?
They will be omitted by the interpreter but still seems bit odd. Is it because the line is too long?

backend/resources/contact.py
```diff
@@ -23,9 +23,8 @@ def post(self):
             reply_to=new_msg["email"],
             recipients=["cfp@codeforpoznan.test"],
         )
-        msg.body = "Nowa wiadomość od {}, nr tel: {} \nTreść:\n {}".format(
-            new_msg["name"], new_msg["phone"], new_msg["content"]
-        )
+        msg.body = f"Nowa wiadomość od {new_msg['name']}, \
```
arturtamborski: I don't want to rule the formatting as it's very opinionated but I'd pull these values out to simpler variables, like so:
```
name, phone, content = new_msg['name'], new_msg['phone'], new_msg['content']
msg.body = f"Nowa wiadomość od {name}, nr tel: {phone} \nTreść:\n {content}"
```
Which seems bit cleaner to me. Maybe the vars shouldn't be stacked in one line too, but that's okayish imo.
Anyway, I'd like to hear the opinion of rest of the team on that.

backend/resources/contact.py
```diff
@@ -23,9 +23,8 @@ def post(self):
             reply_to=new_msg["email"],
```
arturtamborski: You haven't changed the formatting in `subject=` just two lines above, why?

halsing:
Hi,

I didn't put values in the tuples, so I'm suprised. I will chceck as soon
as possible this strange thing. Maybe black library changed it.

In other examples, your way is better than my so I thank you for the
review.

Adrian

sob., 1 lut 2020, 21:45 użytkownik Artur Tamborski <notifications@github.com>
napisał:

halsing:
@stanislawK, now everything should be OK. I changed this string formatting and fixed merge conflict. Also I changed string formatting in new code.

PR #265, opened 2020-03-24
Projects descriptions

MargoWM:
Project descriptions added


OtisRed:
As we discussed, try to paste these data here:
https://github.com/CodeForPoznan/codeforpoznan.pl_v3/blob/master/frontend/src/assets/projects.js

(your PR will still be available when I close it if you would wann copy sth from here)

PR #267, opened 2020-03-25
Wait for migrations

OtisRed:
For some time now we've been talking about the problem with migrations when backend containter stands up quicker than db container. @magul suggested how it should go and...

original

... seems to work :D


OtisRed:
@w1stler maybe you would know what happened here?

```
=================================== FAILURES ===================================
____________________________ test_populate_database ____________________________
script = <flask.cli.ScriptInfo object at 0x7f436d976d30>
    def test_populate_database(script):
        """Test populate_database command."""
        runner = CliRunner()
        result = runner.invoke(populate_database, obj=script)
        app = script._loaded_app
>       assert result.exit_code == 0
E       AssertionError: assert 1 == 0
E        +  where 1 = <Result IntegrityError('(sqlite3.IntegrityError) UNIQUE constraint failed: user.username')>.exit_code
tests/test_commands.py:24: AssertionError
----------------------------- Captured stderr call -----------------------------
```

magul:
Where did you encounter this error @OtisRed ? Is this pull request blocked by this question?

magul:
(BTW, as that's not a scope of this pull request - shouldn't we remove `flask db migrate` from here? as it's responsible for creating new migrations and probably creating them explicitly could be considered beneficial. What do you think @stanislawK ?)

OtisRed:
@magul in reference to the first question it was the built error in travis. I forgot to delete it, as re-run resolved it. Sometimes factoryboy generates the same values in fields with unique constraint and then travis fails, while testing.

PR #275, opened 2020-04-22
Create participant form (fixes #53)

### LINKED ISSUE #53: Add participant form component
Form should contain fields:
 * first name
 * last name
 * email
 * phone
 * github account

OtisRed:
#### Story / Bug id:
#53
#### Description:
Provides the form for creating a new participant and adding it to db

#### Migrations:
N/A

N/A

1. Try to create a new participant with proper data and check if you can add him/her to the hacknight
1. Try to fill the form fields all kind of wrong and check if messages reflect that situation
1. Try to fill the form fields all kind of wrong and check if you can submit the form
1. Try to fill email, github, last_name, slack and phone with the data of an existing user and see if the messages and warnings are user friendly. 
1. Try to send the form without slack and phone (they aren't required)


frontend/src/store/modules/participant.js
```diff
@@ -4,12 +4,16 @@ export default {
...
+    error: null,
+    Participant: null
   },
   getters: {
     getParticipants(state) {
       return state.allParticipants;
     },
+    getEmails(state) {
+      return state.allParticipants.filter(list => list.email);
+    },
```
OtisRed, line 16: @stanislawK @Bosmanfrx if you have a time could you look at these lines? It either renders object that is undefined or doesn't render at all which is why it is undefined (it's also possible that I use wrong the filter() function syntax). Either way I don't get an array of e-mails form the `allParticipants` array of nested json objects as I would want to.

frontend/src/components/Participants.vue
```diff
@@ -0,0 +1,288 @@
...
+        slack: '',
+        phone: ''
+      },
+      warningAlert: false,
+      warningMessage: '',
+      successAlert: false
+    };
+  },
+  created() {
+    this.$store.dispatch('participant/getParticipants');
```
stanislawK, line 115: We are dispatching this action twice: when creating Participants component and HacknightsParticipants component. Since both are child components for Dashboard, maybe it is worth to move that code level up - to dashboard?

frontend/src/components/Participants.vue
```diff
@@ -0,0 +1,288 @@
...
+            user => user.email === value
+          );
+        }
+      },
+      github: {
+        required,
+        githubExists(value) {
+          return !Object.values(this.allParticipants).find(
+            user => user.github === value
+          );
```
stanislawK, line 139: Why we store github urls instead of github username? Right now I am not able to add https:/
/github.com/RozaliaJasica (that github exists in db), but I can add RozaliaJasica. So if we want to keep urls, it could be worth to somehow validate it, or provide some template, to make that field persistent. Or You could just prepopulate that field with 'https://github.com/'

frontend/src/components/Participants.vue
```diff
@@ -0,0 +1,288 @@
...
+  },
+  created() {
+    this.$store.dispatch('participant/getParticipants');
+    setInterval(() => {
+      this.successAlert = false;
+    }, 10000);
+    setInterval(() => {
+      this.warningAlert = false;
+    }, 30000);
+  },
```
stanislawK, line 122: Maybe instead of set intervals running all the time in the background, we could implement that differently ? For example these same behaviour could be achieved using setTimeout function inside methods.

frontend/src/components/Participants.vue
OtisRed: Agreed, but since both would be the children of dashboard should we pass it somehow? It's the unknown territory for me.

frontend/src/components/Participants.vue
OtisRed: Yes I know, but I believe serlizer doesn't accept urls anymore, does it? I remember getting rejects in response when I used url validation.

frontend/src/components/Participants.vue
OtisRed: Seems reasonable. I'll remember reading about it but for some reason I gave up on this. Don't remember why. I'll try again

OtisRed: Ok it worked - changes committed.

frontend/src/components/Participants.vue
stanislawK: There is one store for whole app and You would be able to access participants with `this.$store.dispatch('participant/getParticipants');` same as now. With `this.$store.dispatch('participant/getParticipants')` we pulling participants from backend and "storing" them in central place that every component has access to these same set of data.
So right now we send request to backend when creating Participants component and second time when creating HacknightsParticipants component.

frontend/src/components/Participants.vue
stanislawK: I'm guessing that url could be too long - serializer validates that field with max length 38 chars. So maybe we could just replace input label from github link to e.g. github username or account ?

frontend/src/components/Participants.vue
```diff
@@ -0,0 +1,286 @@
...
+      )
+        this.warningAlert = true;
+      this.warningMessage =
+        'This phone number already exists. Please verify if a person is actually a new paricipant';
+      setTimeout(() => (this.warningAlert = false), 10000);
+    }
+  },
+  computed: {
+    ...mapGetters('participant', ['getError']),
+    ...mapState('participant', {
```
stanislawK, line 215: That solution is working, but as far as I know better practice is to access state with getters, not directly.
This should work e.g.
`...mapGetters('participant', {getError: 'getError', allParticipants: 'getParticipants'})`

frontend/src/components/Participants.vue
OtisRed: Done

frontend/src/components/Participants.vue
OtisRed: Ok. The issue was that RegExp the way as it was written didn't accept signs like `/`. 

At the same time after discussing with @stanislawK we agreed that there is no reason to store URLs rather than bland usernames so I leave it as it is. 

I also added a method, which deletes "https://github.com/" automatically in case someone would want to copy/paste URL from github

frontend/src/components/Participants.vue
OtisRed: Done

frontend/src/components/Participants.vue
```diff
@@ -0,0 +1,303 @@
...
+      return errors;
+    },
+    phoneErrors() {
+      const errors = [];
+
+      if (!this.$v.form.phone.$dirty) return errors;
+      !this.$v.form.phone.numeric &&
+        errors.push('You can only type digits here');
+      !this.$v.form.phone.minLength &&
+        errors.push('Type correct number ie. 111-222-333');
```
stanislawK, line 287: Minor thing - correct format is 111222333 actual, so that hint is a bit misleading.

OtisRed: I corrected the messages. 

I thought about introducing mask, but:
1) it seems it's been[ deprecated in vuetify above 2.0](https://stackoverflow.com/questions/47857673/how-to-use-mask-in-vuetify-text-field) and would require external package like [this](https://vuejs-tips.github.io/vue-the-mask/);
2) Such a mask, changes the value of the field and I'm not sure if we want to store numbers like this 111-222-333 (with dashes). Of course we could replace it before sending, but at this point it seems too much of a hustle.

If you feel like we could do it anyway please create a separate task for introducing vue-the-mask or some other package and replace it both in this form and contact form on the landing page (mask stopped working there).

stanislawK (approved):

Issue #233, opened 2020-01-08
 AWS deployment 


MargoWM:
that requires:
* database
* migration lambda
* serverless API
* frontend asset
* CloudFront

PR #283, opened 2020-05-24
[233] AWS deployment (backend)

arturtamborski:
#### Story / Bug id:
[233](https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/233)

#### Description:
This pull request will add build-and-deploy step of the project to codeforpoznan's AWS Lambda.
It will be launched at every merge to master.

One notable change that unfortunately I do have to make in order to keep the complexity at minimum is to move all API endpoints to sub-url - `/api/`. I hope that this is not too much of a problem.

This requirement is quirky in its nature but in short - forwarding requests to `/` (root) is tricky because of pattern matching at origin level. There are some ways to overcome that but that would require a second lambda (@ Edge) in CF distribution to match & forward the requests in appropriate way.
I'm happy to explain this in depth and show the problem via screenshare if that's needed.

Please, note that pull request is a draft. I need to get the secret in travis.yml filled (@magul?) first, then the infra PR needs to be applied (https://github.com/CodeForPoznan/Infrastructure/pull/61) and then this PR will need to be tested with actual infra underneath.
But anyway, theoretically this _should_ work, just as the one in pah-fm works :)

Also, this line https://github.com/CodeForPoznan/codeforpoznan.pl_v3/blob/5927952e81703b0721101a1aa21872df1d0ae215/backend/resources/contact.py#L25 needs to be parametrized, preferably with some logic based on environment variables (my proposal is to use `BASE_URL`), because it will differ for actual production environment and the dev one laid out here.

#### Migrations:
N/A

`serverless-wsgi==1.7.4`

Confirm that all endpoints were safely moved over to `/api/`.
Confirm that the bash script does its job.
Confirm that travis runs all tests as it did before and doesn't have any errors (build should launch, run, and succeed).


arturtamborski:
Update: frontend work is done, was split from this PR to another one - https://github.com/CodeForPoznan/codeforpoznan.pl_v3/pull/286 and later on fixed a bit here https://github.com/CodeForPoznan/codeforpoznan.pl_v3/pull/289.

We're still blocked by https://github.com/CodeForPoznan/Infrastructure/pull/61 and the email line mentioned above.

arturtamborski:
Another Update: Infra ticket is ready to be reviewed (and it was also applied) so I moved forward with this PR.

I'm happy to announce that the lambdas are in fact deployed and are responding to requests, but there is still one issue with imports that will need to be solved in order to merge and close this one.

arturtamborski:
I also have to update the travis encoded secret because that changed when I moved IAM user to `dev_` in infra. I'm counting on your wizard encryption skills here, @magul.

OtisRed:
@magul you are wanted to generate secrets here :)

magul:
@arturtamborski #294 - we need to merge it and you can backmerge it then.

arturtamborski:
I'm closing this one because it's way to big and too hard for a solid review. I'll create few smaller pull requests instead.

Issue #272, opened 2020-04-15
Redirect to the dashboard after login


stanislawK:
User should be redirected to the dashboard after successfully logged in

PR #293, opened 2020-07-08
[272] Redirect after login

stanislawK:
#### Story / Bug id:
[272](https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/272)

#### Description:
User after successfully logged in, would be redirected to dashboard

#### Migrations:
N/A

"vue-router": "^3.3.4"

Navigate to http://localhost:8080/login, login with valid username and password. After that redirection to http://localhost:8080/dashboard should be triggered. Additionally It could be required to rebuild frontend with newer version of vue-router.


frontend/src/store/modules/auth.js
```diff
@@ -34,6 +38,7 @@ export default {
         localStorage.setItem('token', token);
         localStorage.setItem('refresh_token', refresh_token);
         axios.defaults.headers.common['Authorization'] = 'Bearer ' + token;
+        commit('setToken', token).then(router.push('/dashboard'));
```
OtisRed, line 41: Not sure what it is about, but the problem seems to be traced to here.

Zrzut ekranu z 2020-07-29 22-32-19

Issue #300, opened 2020-08-19
DB model for teams


MargoWM:
Fields:
**descriptions
**project name
**participants
**project URL

PR #304, opened 2020-08-26
[300] Team DB model

stanislawK:
#### Story / Bug id:
[300](https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/300)

#### Description:
DB model for teams

#### Migrations:
8c7cbf0f76c6

N/A

New models should be created when starting up docker. To verify that run:
`docker-compose exec db psql -U cfp_v3 -c "\dt"`
team and participant_team should be present in list of tables


OtisRed:
DO NOT MERGE BEFORE #306 - this PR is perfect to test migration in AWS Lambda

backend/migrations/versions/8c7cbf0f76c6_.py
```diff
@@ -0,0 +1,45 @@
+"""empty message
```
w1stler, line 1: This could be filled. ;-)

w1stler (approved):

OtisRed:
@arturtamborski how do we test if this migration works on AWS?

arturtamborski:
> @arturtamborski how do we test if this migration works on AWS?


Quick way:
Click on build icon near latest commit, then details and the link:
image


Or...


1. Go to list of commits:
https://github.com/CodeForPoznan/codeforpoznan.pl_v3/commits/master

2. Find the merge commit:
https://github.com/CodeForPoznan/codeforpoznan.pl_v3/commit/c6a5941e679f3ac739df3d16ff7ff3b772637234

3. Click on the green OK sign (or in case of a failure, red X):
image

4. Click on details, this will open up Actions view in github:
https://github.com/CodeForPoznan/codeforpoznan.pl_v3/runs/1221966304

5. Click on "The build" link:
https://travis-ci.com/github/CodeForPoznan/codeforpoznan.pl_v3/builds/188649012

Now, scroll down to the post-build stuff:
image

Unroll "Deploying application" and read it, to figure out if it passed.
**Please note that the build might be green but the deployment may still fail!** which is the case this time. It's important to examine the logs during important changes (such as merging PR with migrations) to confirm that the deployment succeeded.

Unfortunately, the migrations are very fragile to wrap up in a bulletproof way hence the need for checking deployment result when someone merges migrations. For other cases the deployment should be a no-brainer.

In our case the output is as follows:
```
run migrations
show migration output
Running migrate

INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
ERROR [root] Error: Can't locate revision identified by '849ece77fb7c'

Done. Your build exited with 0.
```

So the migration were ran but they failed with `Can't locate revision identified by '849ece77fb7c'`.
I don't know why yet, so I'll check that right now and correct the migration but the general flow looks as described.

arturtamborski:
After some testing I think it failed because I left some incorrect values in `alembic_version` table during last testing session? I'm not sure, anyway, I cleared all data (including the dummy records), restarted the migrations and it all now looks fine and dandy.
The latest version of dev.codeforpoznan.pl has the `teams` table (introduced in this PR). I think it's ready for further development.

I don't expect any more problems with migrations :)

PR #306, opened 2020-09-10
[233] Add lambda handler and deployment steps

arturtamborski:
#### Story / Bug id:
#233 

#### Description:
I finally implemented it!

#### Migrations:
N/A

new:
serverless_wsgi==1.7.5
upgrade:
psycopg2-binary==2.8.5
Flask-SQLAlchemy==2.4.4

go to [dev.codeforpoznan.pl](https://dev.codeforpoznan.pl/login) and try it out! :D 
credentials: `admin:pass123`


OtisRed:
@arturtamborski I must have done sth wrong :D

Zrzut ekranu z 2020-09-11 21-14-28

arturtamborski:
Yea, to be fair I'm still working on the details here. Lambda itself works fine, migrations are running OK, the schema was created and you can run the app with remote DB just fine, so it's _nearly_ finished, but I still have to tweak the frontend code a bit.

For some odd reason axios is pumping all requests to `0.0.0.0:5000` instead of the proper host, so that's on my mind right now.
Other than that I sure think it's finally ready! :D

arturtamborski:
Update: axios is now working fine which makes this PR ready for a review.

List of things to do / check in future:
1. **e-mails**. I have no idea if that piece of code in emails will be enough in terms of the app logic. We still have to set up SES for mail forwarding but that's a story for another time.
2. **migrations**. We have one handy PR that will let us test the migrations swiftly and I'm sure they will work fine but as always in these cases, this assumption is still to be confirmed.
3. **caching**. Something is wrong with the way API gateway returns responses and I'm blaming caching for that. I'll check that as my first priority because it severely affects the way app works.

But taking things aside - its working just fine :) The login works, frontend works, backend works, migrations are working and its a good (although _very_ late) start! :)

jacekkalbarczyk (approved):
Works fine, nothing to add here.

PR #307, opened 2020-09-10
[233] Prefix all API urls with /api/

arturtamborski:
#### Story / Bug id:
#233 

#### Description:
I prefixed all API urls with `/api/`. This is needed for Lambda.

make test
and click stuff in panel, everything should work as if this PR never happened.


OtisRed (approved):
I tested everything I thought of automatically and manually and it seems to work fine

PR #319, opened 2020-10-21
Renew slack link

OtisRed:
#### Story / Bug id:
Old slack invite link is expired. I added a new one.


OtisRed:
Something was wrong with git here

PR #320, opened 2020-10-21
Renew slack invite

OtisRed:
#### Story / Bug id:
I've renewed slack invite link


OtisRed:
Sorry for self-approving but its really not a real change here

Issue #292, opened 2020-07-01
Fix styles and check for existing data in Modal


agatakozinska:
#### Subject of the issue:
There are differences between modals, view for them should be consistent. Also there is no checking if some data exist.

#### Actual behaviour:
*In one project all headings are displayed in uppercase, in another they are normal. 
*Also for e.g. `{{ selectedProject.partner }}` there is no checking if actually partner exists. I don't know Vue but in general, in JS, you should check if some data exists and if not a whole div shouldn't be displayed, because there is no need to show even a heading with no data provided
![83687075_562836867956139_5992232061940397045_n](https://user-images.githubusercontent.com/25872895/86296307-ca632480-bbf8-11ea-83e9-a90c19161265.jpg)
![106506253_3117334408352278_222767838618080432_n](https://user-images.githubusercontent.com/25872895/86296311-cb945180-bbf8-11ea-8520-5a0c0152b4e3.jpg)

#### Expected behaviour:
All heading in every modal should have the same style for text.
About data, I know that no every project, for now, have description, src for repo etc. but I guess not every project will have a partner and that section should be shown only if it has one.

PR #324, opened 2020-11-11
Refactor modal content (fixes #291 and @292) 

### LINKED ISSUE #291: Fix header in Modal
**Subject of the issue:**
In mobile view, there is a problem with header in Modal component.

**Your environment:**
mobiles, resolution < 500px

**Actual behaviour:**
When the title of a project is longer it breaks oddly also the close button is not fixed in one place, sometimes is covered by title.

83687075_562836867956139_5992232061940397045_n
106506253_3117334408352278_222767838618080432_n

**Expected behavior:**
One word shouldn't be divided between two lines (word-break shouldn't be set on `break-all` use `break-word` instead).
The button is displayed in the top, right corner and is not covered by title.

OtisRed:
#### Story / Bug id:
#291 and #292 

#### Description:
It was easier for me conceptually to go through both those tickets at the same time. Most important changes incude:

1. Title breaks by words in case of overlay - it was breaking by letters before.  
1. I introduced h2 and h3 tags to structure the content for SEO reasons.
1. In mobile view I moved the close button to the bottom-right corner. It made a lil bit more space for the title on smaller devices and made it easier to use close button in mobile view (it's closer to a thumb). 
1. I centered all the content. Seems more intuitive this way.
1. All the buttons are rendered conditionally based on the condition that there actually is something to show.
1. "Partner" and "stack" sections are also rendered conditionally. 
1. I recomposed partner/s section, because we had long text buttons that wouldn't break in case of an overlay.
1. I dropped v-btn, v-icon and v-card - I styled them with CSS and moved some scoped classes to global.scss

#### Migrations:
I nested partner partners' attributes in the partner object, which is a list now.

N/A

1. Run in docker
1. Check visually - especially on smaller screens
1. Check if conditional rendering works properly


frontend/src/components/ModalContent.vue
```diff
@@ -1,119 +1,116 @@
...
-          >
-            <v-icon size="3.5rem" color="white">close</v-icon>
-          </v-btn>
-        </v-card-actions>
-      </v-row>
+    <v-card-title class="container-title unified-padding">
+      <h1>{{ selectedProject.name }}</h1>
+      <div>
+        <v-btn class="hidden-xs-only" rounded icon large @click="onClick()">
+          <v-icon size="3.5rem" color="white">close</v-icon>
```
agatakozinska, line 7: would be better to pass it as some variables not hardcoded like that, if not pass color as hex (#ffffff) not `white`
you can try this approach: 
https://stackoverflow.com/a/61378302

frontend/src/components/ModalContent.vue
```diff
@@ -1,119 +1,116 @@
...
-            class="buttons"
-            text
-            rounded
-            large
-            :href="selectedProject.websiteLink"
-            target="_blank"
+    <v-card-actions class="buttons-list unified-padding">
+      <div v-show="selectedProject.licensePage !== ''">
+        <v-btn text rounded :href="selectedProject.licensePage" target="_blank">
+          <v-icon>far fa-copyright</v-icon>
```
agatakozinska, line 14: why not to use https://vuetifyjs.com/en/features/icons/#using-custom-icons  and create component with icons? this way you write it once and then use it in a whole project
also I'd add some accessibility here like `aria-label`

frontend/src/components/ModalContent.vue
```diff
@@ -1,119 +1,116 @@
...
-          >
-            <v-icon>fas fa-handshake</v-icon>
-            <p class="buttons--text">Strona partnera</p>
-          </v-btn>
-        </v-row>
+        </v-btn>
+      </div>
+      <div v-show="selectedProject.githubLink !== ''">
+        <v-btn text rounded :href="selectedProject.githubLink" target="_blank">
+          <v-icon>fab fa-github</v-icon>
```
agatakozinska, line 22: ditto

frontend/src/components/ModalContent.vue
```diff
@@ -1,119 +1,116 @@
...
-      </v-row>
-    </div>
-    <div class="text-list">
-      <v-row>
-        <p class="content-black">{{ selectedProject.description }}</p>
-      </v-row>
+    <v-card-text class="unified-padding">
+      <p class="text-center">{{ selectedProject.description }}</p>
+    </v-card-text>
+    <div v-show="countPartners >> 0">
```
agatakozinska, line 36: is `>>` it ok? not just `>` ?

frontend/src/components/ModalContent.vue
```diff
@@ -1,119 +1,116 @@
...
+          v-for="partners in selectedProject.partner"
+          :key="partners.name"
+        >
+          <v-col class="partner-item" cols="12" sm="4">
+            {{ partners.name }}
+          </v-col>
+          <v-col class="partner-item" cols="12" sm="3">
+            <v-btn class="buttons-chips" text :href="partners.link" rounded>
+              <span class="buttons-text">Poznaj</span>
+              <v-icon>fas fa-hands-helping</v-icon>
```
agatakozinska, line 61: ditto

frontend/src/components/ModalContent.vue
```diff
@@ -125,59 +122,83 @@ a {
...
+    margin-right: 0;
+  }
 }
 
-.text-list {
-  margin: 0.25rem 3rem;
+h2 {
+  font-family: $font-content;
+  font-size: 1.5rem;
+  color: black;
```
agatakozinska: use variable

frontend/src/components/ModalContent.vue
```diff
@@ -125,59 +122,83 @@ a {
...
+  word-break: break-word;
+  @media only screen and (max-width: $phone) {
+    text-align: center;
+  }
 }
 
-.buttons {
-  margin: 1rem 0.5rem 1rem 0;
+.buttons-close {
+  background-color: $blue !important;
```
agatakozinska: is `!important` needed? I guess you can add another class to the element and then you don't have to use such an overwriting

frontend/src/components/ModalContent.vue
```diff
@@ -125,59 +122,83 @@ a {
...
 }
 
-.title-header {
-  background: $blue;
+.partner-item {
+  display: flex;
+  justify-content: center;
+  text-align: center;
+  @media only screen and (max-width: $phone) {
+    padding: 2% 10%;
```
agatakozinska, line 189: it's not easy to predict how it will behave on different devices (resolutions), if you use rem I would stay with that, not mixed

frontend/src/components/ModalContent.vue
```diff
@@ -125,59 +122,83 @@ a {
...
+  }
+}
+
+.unified-padding {
+  padding: 0.5rem 1.5rem !important;
```
agatakozinska: same as above - add another class, `!imporant` should be the laaaaaaaaaaast thing you can think about xd

agatakozinska (changes requested):

frontend/src/components/ModalContent.vue
```diff
@@ -1,119 +1,116 @@
...
-            large
-            right
-            @click="onClick()"
-          >
-            <v-icon size="3.5rem" color="white">close</v-icon>
-          </v-btn>
-        </v-card-actions>
-      </v-row>
+    <v-card-title class="container-title unified-padding">
+      <h1>{{ selectedProject.name }}</h1>
```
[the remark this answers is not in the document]
OtisRed: Given our discussion of h1 tags we should decice whether using h1 here is appropriate at all. Whatever the class should be I wouldn't argue moving at least a port of it to the global class. Will keep you posted on the adjustments.

frontend/src/components/ModalContent.vue
OtisRed: Ok. so the options in the link regard the situation when I actually want to change sth about default vuetify settings. Given that I realize how fucked up it is to have part of the styling through vuetify props and others through regular classes is there actually a reason to try to overwrite that? I mean veutfy is what it is, but isn't it overkill to write css just to override the default settings to the same effect? Please straight me out if I'm wrong.

frontend/src/components/ModalContent.vue
OtisRed: That seems awsome. I'll try to implement that.

frontend/src/components/ModalContent.vue
OtisRed: You're right. For some reason webstorm was correcting me on a singular use, but actually it seems to work fine. FYI VSC doesn't claim that kind of problem :D

frontend/src/components/ModalContent.vue
OtisRed: Well... we didn't have global black defined :D I'll work on that => promise (I realize that promise isn't usua;;y what you return in a function => just an inside joke :P )

frontend/src/components/ModalContent.vue
OtisRed: Yes, this is the kind of moment when not using important results in nothing. So either I can use in-line prop where I cannot use $var and need to repeat $blue var's hex or I can use important to overrride whatever vuetify understands. I didn't figure out how to pass global $blue into html part of code

frontend/src/components/ModalContent.vue
OtisRed: Good point. However if it was you to consider some kind of understand what measure would you use? Is rem the best choice?

frontend/src/components/ModalContent.vue
OtisRed: Really it was. I usually use !importnat when I feel helpless with vuetify :P However this change had been override and I will actually try to replace vuetify for clear vue in the whole component. We'll see if that's needed then.

frontend/src/components/ModalContent.vue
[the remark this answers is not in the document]
OtisRed: I removed h1 tag to reserve it for singular use only - although multiple h1 tags don't matter anymore in html 5, because web crawlers use sections to understand the structure in more modular way (e.g. [Google](https://www.youtube.com/watch?v=WsgrSxCmMbM)) people still believe that it is the best practice. Basically the argument is that with the singular use you don't lose anything but you ay gain sth in particular cases. 

So I used h2 and h3 tags instead - both went to global classes.

frontend/src/components/ModalContent.vue
OtisRed: **Final changes:**
1. I dropped v-btn and v-icon altogether (in this case vuetify does more harm than good)
2. I created global classes for two kinds of buttons which use globally defined colors

frontend/src/components/ModalContent.vue
OtisRed: **Final changes:**
Global store for icons seems to be a really good idea but when I started playing with it both through Vuetify and Font Awsome I couldn't find the way that actually save code. To the contrary, it did introduce another layer of abstraction which doesn't seem that necessary at this point. For now I settled with the basic local use of FA with <i> tag. 

I still believe global store for icons is a good idea but I think it goes beyond the scope of this task to do it right. I will create a new issue for that.

frontend/src/components/ModalContent.vue
OtisRed: ditto

frontend/src/components/ModalContent.vue
OtisRed: ditto

frontend/src/components/ModalContent.vue
OtisRed: **Final changes:**
That's done

frontend/src/components/ModalContent.vue
OtisRed: **Final changes:**
I dropped v-btn altogether so that's not an issue no more

frontend/src/components/ModalContent.vue
OtisRed: **Final changes:**
U used REMs both for desktop and mobile view

frontend/src/components/ModalContent.vue
OtisRed: **Final changes:**
I dropped using vuetify's v-card to style the container so it's not a problem anymore

stanislawK (approved):

PR #342, opened 2021-01-06
Fixing modal recov (WIP)

OtisRed:
#### Story / Bug id:
Chekout for #324 PR. I created a fresh branch and copy/pasted code by hand so any interventions with packages don't echo to create bugs on the backend. Problem must stem from sth else


OtisRed:
It was created just for the sake of checking sth with the build

PR #345, opened 2021-01-20
4 import attendance list from google sheets

w1stler:
#### Story / Bug id:
4

#### Description:
Download csv from Gdrive and put into backend/temp folder. It's one-time thing. 

#### Migrations:
N/A

N/A

N/A


backend/commands/import_attendance_list.py
```diff
@@ -0,0 +1,80 @@
...
+
+        list_of_participants.append(users_dict)
+        users_dict = {}
+
+    # create hacknights
+    for date in dates:
+        splitted_date = date.split(".")
+        formatted_date = f"{splitted_date[2]}-{splitted_date[1]}-{splitted_date[0]}"
+        hacknight = Hacknight(date=formatted_date)
+        db.session.add(hacknight)
```
arturtamborski, line 50: bit more obvious imo, though I'm not sure if it should be `year-month-day` or `day-month-year`).
```suggestion
        year, month, day = date.split(".")
        hacknight = Hacknight(date=f"{year}-{month}-{day}")
        db.session.add(hacknight)
        
```

backend/commands/import_attendance_list.py
```diff
@@ -0,0 +1,80 @@
...
+        if participant["github"] == "":
+            participant["github"] = f"unkown_github_{idx}"
+        new_participant = Participant(
+            id=participant["id"],
+            github=participant["github"],
+            first_name=participant["first_name"],
+            last_name=participant["last_name"],
+            email=participant["email"],
+            phone=participant["phone"],
+        )
```
arturtamborski, line 67:
```suggestion
        new_participant = Participant(**participant)
```
:)

backend/commands/import_attendance_list.py
```diff
@@ -0,0 +1,80 @@
...
+        db.session.add(hacknight)
+    db.session.commit()
+    click.echo(f"Imported {len(dates)} hacknights")
+
+    # create participants
+    for idx, participant in enumerate(list_of_participants):
+        if participant["email"] == "":
+            participant["email"] = f"unkown_email_{idx}"
+        if participant["github"] == "":
+            participant["github"] = f"unkown_github_{idx}"
```
arturtamborski, line 59:
```suggestion
        participant["email"] = participant.get("email", f"unknown_email_{idx}")
        participant["github"] = participant.get("github", f"unknown_github_{idx}")
```

Two things here:
- typo `unkow` -> `unknown`
- bit simpler approach (some would say - a matter of style)

backend/commands/import_attendance_list.py
```diff
@@ -0,0 +1,80 @@
...
+            email=participant["email"],
+            phone=participant["phone"],
+        )
+        db.session.add(new_participant)
+    db.session.commit()
+    click.echo(f"Imported {len(list_of_participants)} participants")
+
+    # assign attendance list
+    for idx, row in enumerate(rows, 1):
+        for index, element in enumerate(row[5:], 1):
```
arturtamborski, line 74: `idx` and `index` aren't very clear. How about `participant_idx` and `hacknight_idx`?

backend/commands/import_attendance_list.py
```diff
@@ -0,0 +1,80 @@
...
+            participant = Participant.query.get(idx)
+            if element == "1":
```
arturtamborski, line 76: this might be a bit too strict, in case we get a human error (like a space in the cell, "`1 `").
```suggestion
            if "1" in element:
```

docker-compose.yml
```diff
@@ -38,3 +38,5 @@ services:
       POSTGRES_DB: cfp_v3
       POSTGRES_USER: cfp_v3
       POSTGRES_PASSWORD: cfp_v3
+    ports:
+    - "5432:5432"
```
arturtamborski, line 42: big thumbs up for adding quotes! not a problem here, but yaml is very touchy in some places (see https://docs.docker.com/compose/compose-file/compose-file-v3/#ports for more details, not a problem here, but still, it's a good practise).

backend/commands/import_attendance_list.py
arturtamborski: another thing - I'm not so sure if querying it by idx is safe. I don't know how this file looks like but if it's relying on `IDs` to be created incrementally (`1, 2, 3, ...`) then it might not work like so all the time. There can be an ID with out of order number in the DB already.
Probably not an issue for now, and this importer won't be used too much, but still, a heads up from me.

backend/commands/import_attendance_list.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # get dates of historical hacknights
+    dates = rows[0][4:]
+    click.echo(f"Created list of dates: {dates}")
+
+    # skip header
+    rows = rows[1:]
+    # prepare users' data
+    for item in rows:
+        _id, github, name, phone, email, *values = item
```
arturtamborski, line 31:
```suggestion
        _id, github, name, phone, email, *_ = item
```

*incredible nitpick* :P

backend/commands/import_attendance_list.py
```diff
@@ -0,0 +1,80 @@
...
+        first_name = name.split()[0]
+        last_name = name.split()[1]
```
arturtamborski, line 33: this won't work as expected if someone has second name in the table, or double last name. Not sure if it's a problem.

arturtamborski (approved):
Overall very crisp and clear import code. All of my notes are regarding code style, most of these are nitpicks, so you can ignore them at will.

backend/commands/import_attendance_list.py
```diff
@@ -0,0 +1,80 @@
...
+@click.command()
+@with_appcontext
+def import_attendance_list():
+    """Import attendance list."""
+    rows = []
+    list_of_participants = []
+    users_dict = {}
+
+    # open file from exported csv
+    with open("./temp/list.csv") as csv_file:
```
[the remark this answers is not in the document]
arturtamborski: we had a discussion around that part - tldr: it's not worth it.
Passing the argument trough make is bit odd, there are two ways you can do it and both kinda suck in their own way. The import itself is supposed to be ran only once so this doesn't really justify it.

arturtamborski (approved):

backend/commands/import_attendance_list.py
w1stler: Simpler approach doesn't work in this form - it should be `==` I guess.

arturtamborski (approved):

PR #363, opened 2021-03-11
Local setup with remote db

arturtamborski:
#### Story / Bug id:
None

#### Description:
It's a short no-op change for easier set-up and connection to remote DB when testing the app.

#### Migrations:
N/A

N/A

ask for secrets and ssh config and test the README notes.

### DIFF
--- a/.envrc.example
+++ b/.envrc.example
@@ -4,3 +4,4 @@ export DB_PORT=5432
 export DB_NAME=cfp_v3
 export DB_USER=cfp_v3
 export DB_PASSWORD=cfp_v3
+
--- a/.envrc.example
+++ b/.envrc.example
@@ -1,0 +1,6 @@
+export SECRET_KEY=codeforpoznan
+export DB_HOST=db
+export DB_PORT=5432
+export DB_NAME=cfp_v3
+export DB_USER=cfp_v3
+export DB_PASSWORD=cfp_v3
--- a/README.md
+++ b/README.md
@@ -5,9 +5,11 @@
 Next version of Code for Poznan's website.
 
 ## Development environment
+
 We use Docker for development purposes. To start working, your computer needs to support it.
 
 ### Commands
+
 | command      | what it does   |
 |--------------|----------------|
 | `make start` | starts project |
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -16,7 +18,80 @@ We use Docker for development purposes. To start working, your computer needs to
...
+docker-compose exec backend bash
+root@a68d3611d0e7:/cfp_v3/backend# flask shell
+Python 3.8.8 (default, Feb 19 2021, 17:55:44)
+[GCC 8.3.0] on linux
+App: backend.app [development]
+Instance: /cfp_v3/instance
+>>>
+>>> from backend.models import User
+>>> User.query.all()
+[<User 1>, <User 2>, <User 3>, <User 4>]
```
OtisRed, line 90: Is it the result or command?

OtisRed (commented):
Few readability tips from noob :)

README.md
```diff
@@ -16,7 +18,80 @@ We use Docker for development purposes. To start working, your computer needs to
...
+make start
+```
+
+How to confirm that it works:
+
+```bash
+$ make bash
+docker-compose exec backend bash
+root@a68d3611d0e7:/cfp_v3/backend# env | grep DB
+DB_PASSWORD=supersecretvalueasdasdadas
```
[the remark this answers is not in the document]
arturtamborski: Oh, see it's not for pasting - this should only show you that these passwords were in fact loaded up correctly.

README.md
```diff
@@ -16,7 +18,80 @@ We use Docker for development purposes. To start working, your computer needs to
...
+DB_PORT=55432
+[...]
```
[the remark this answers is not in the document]
arturtamborski: [more output here], will add comment.

README.md
arturtamborski: yes, that's a list of objects fetched from database. Will add description.

.envrc.example
```diff
@@ -0,0 +1,6 @@
+export SECRET_KEY=codeforpoznan
+export DB_HOST=db
+export DB_PORT=5432
+export DB_NAME=cfp_v3
+export DB_USER=cfp_v3
+export DB_PASSWORD=cfp_v3
```
OtisRed, line 6: I know it's an example but come on :D

README.md
```diff
@@ -16,7 +18,169 @@ We use Docker for development purposes. To start working, your computer needs to
...
+```bash
+# run this in terminal, in the project directory
+cp .envrc.example .envrc
+```
+
+</br>
+
+#### 2. Edit the `.envrc` file using your favourite text editor
+
+Replace the secret values (passwords and secrets).
```
OtisRed, line 59: What is a secret in .envrc file? There is no field secret isn't it?

OtisRed (approved):
Now I get it. I left two more points I you feel like explaining them :) but it's ok now I think :)

.envrc.example
arturtamborski: no, that's on purpose. If someone by mistake or for any other reason will copy the `envrc.example` to `.envrc` it will not break his environment, because the default variables are the same as defined in `docker-compose.yaml`.
I could put something else in here, but that would potentially break env if someone would follow the procedure from readme without replacing the secrets with something more realistic.

README.md
arturtamborski: Oh, you got me there. I used `secrets` to refer to basically every key in `.envrc.example`. I consider both passwords as well as usernames to be 'secret' - as in - "it shouldn't be known, even if the value is not surprising".

.envrc.example
arturtamborski: In other words, it's a safeguard for a hypotetical user who would follow the whole procedure but failed to obtain the secrets from admins. If these values would be different then he would probably end up with broken DB connection.

Issue #120, opened 2019-07-17
Configure e-mail box for SES 


OtisRed:
**General description**
1. Given that we deploy our app on AWS we need to use native services to get e-mails from the website. 
1. Also it would be nice to catch e-mails when app is run in docker to test delivery locally. 

**Technical description**
Use Simple Email Service in AWS to send and receive e-mails

1. Create a bucket for e-mails
1. Create a queque in SES
1. Link the two together 
1. Create policies 

**Support documentation**
https://aws.amazon.com/premiumsupport/knowledge-center/ses-receive-inbound-emails/

OtisRed:
@jacekkalbarczyk Jacek can  you take it?

OtisRed:
**Possible solutions**
1. Using SES shoudln't be an issue given that we already have it in the infrastructure
2. One of the solutions is to import SES from Boto: https://www.learnaws.org/2020/12/18/aws-ses-boto3-guide/ 
3. Bucket is needed only if we assume we want to test sending e-mails - if don't we can disregard it

arturtamborski:
Some explanation of the S3 bucket thingy - originally I had an idea where we'd use s3 bucket for storing mails from `dev` domain for debugging, but actually send the mails from the `prod` domain. This would allow us to test the mail form without spamming `hello@` with debug mails. Technically it's not required, because we can hardcode sending mails to `hello@` in both domains and just deal with the spam in other way (gmail filters?), but yeah, it's not required.

The only required thing here to do is to enable SES endpoint so that we can send mail from backend, that's all.

Now, on how to actually send the mail there are three approaches:
1. Send it via raw HTTP requests - not recommended, tricky to get right
2. Send mails with `boto3` (aws library) - super easy to get working
3. Send mails with `flask-mail` via SMTP protocol - more complicated than [2], still doable and useful if we would want to use other SMTP provider.

Last thing I want to talk about is debugging the mails from `dev` domain. It's surprisingly tricky, because access to `hello@` is restricted and that's where the mails should go so it's either spamming your mailbox or creating temp mails or... we could use a service focused on testing the mails! :) https://ethereal.email/
They offer a free way of sending mails to their server which will then give us a pretty link where we can inspect the resulting message without it being actually sent (by mistake or intentionally). I'd want to enable this service for any mail-sending situations where it's _not_ production. In other words, I'd like to use this mail-capturing service for localhost, dev or any other environment which is not production.

What do you think?

PR #375, opened 2021-04-16
120 send email

arturtamborski:
#### Story / Bug id:
https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/120

#### Description:
Introduced mail sending code with testing features :)

#### Migrations:
N/A

requests, needed for easy request making

```
make start && make logs
```

fill the form on main page and watch the logs for message with link to `ethernal.mail` :)

NOTE: This is a draft, still needs few enhancements and more importantly - it's blocked by the infrastructure ticket!

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -14,7 +14,7 @@ COPY Pipfile Pipfile.lock ./
 
 RUN pip install --upgrade pip \
     && pip install pipenv==v2020.11.15 \
-    && pipenv install --system --deploy
+    && pipenv install --dev --system --deploy --verbose
 
 ADD . /cfp_v3/backend
 
--- a/backend/app.py
+++ b/backend/app.py
@@ -1,3 +1,5 @@
+import os
+
 from flask import Flask
 from flask_cors import CORS
 
@@ -25,7 +27,8 @@ from backend.resources.team import TeamDetails, TeamList, TeamMembers
 def create_app():
     """Application factory function"""
     app = Flask(__name__)
-    app.config.from_object("backend.config.DevelopmentConfig")
+    env = os.environ["FLASK_ENV"].title()
+    app.config.from_object(f"backend.config.{env}Config")
     app.cli.add_command(populate_database)
     app.cli.add_command(remove_expired_tokens)
     app.cli.add_command(import_attendance_list)
--- a/backend/config.py
+++ b/backend/config.py
@@ -1,32 +1,85 @@
 import os
+import logging
+
+import requests
 
 
-class Config(object):
-    DEBUG = False
-    TESTING = False
+class Config:
+    """
+    Default config context.
+    """
 
-
-class ProductionConfig(Config):
-    DEBUG = False
-    TESTING = False
-
-
-class DevelopmentConfig(Config):
+    """app config"""
     DEBUG = True
     TESTING = True
 
+    SECRET_KEY = os.environ["SECRET_KEY"]
+    BASE_URL = os.environ["BASE_URL"]
+
     """db config"""
-    DB_USER = os.environ.get("DB_USER")
-    DB_PASSWORD = os.environ.get("DB_PASSWORD")
-    DB_HOST = os.environ.get("DB_HOST")
-    DB_PORT = os.environ.get("DB_PORT") or 5432
-    DB_NAME = os.environ.get("DB_NAME")
+    DB_HOST = os.environ["DB_HOST"]
+    DB_PORT = os.environ["DB_PORT"] or 5432
+    DB_NAME = os.environ["DB_NAME"]
+    DB_USER = os.environ["DB_USER"]
+    DB_PASSWORD = os.environ["DB_PASSWORD"]
 
     SQLALCHEMY_DATABASE_URI = (
         f"postgresql://{DB_USER}:{DB_PASSWORD}@{DB_HOST}:{DB_PORT}/{DB_NAME}"
     )
     SQLALCHEMY_ENGINE_OPTIONS = {"pool_recycle": 299}
     SQLALCHEMY_TRACK_MODIFICATIONS = False
-    SECRET_KEY = os.environ.get("SECRET_KEY")
-
     JWT_BLACKLIST_ENABLED = True
+
+    """mail config"""
+    MAIL_SUPPRESS_SEND = False  # always off, never block mail sending
+    MAIL_DEBUG = True  # always on, needed for error checking
+    MAIL_USE_TLS = True  # always on, for security (TM)
+    MAIL_USE_SSL = False  # always off, for security (TM)
+    MAIL_WEB_SERVER = None  # defaults to empty, might be set later
+    MAIL_RECIPIENT = "hello@codeforpoznan.pl"
+
+    MAIL_SERVER = os.environ.get("MAIL_SERVER")
+    MAIL_PORT = os.environ.get("MAIL_PORT")
+    MAIL_USERNAME = os.environ.get("MAIL_USERNAME")
+    MAIL_PASSWORD = os.environ.get("MAIL_PASSWORD")
+
+    # empty MAIL_* vars are a hint that we should use the temporary account
+    if not MAIL_SERVER:
+        json = {"requestor": "nodemailer", "version": "6.5.0"}
+        resp = requests.post("https://api.nodemailer.com/user", json=json)
+        if resp.status_code == 200:
+            data = resp.json()
+            MAIL_SERVER = data["smtp"]["host"]
+            MAIL_PORT = data["smtp"]["port"]
+            MAIL_USERNAME = data["user"]
+            MAIL_PASSWORD = data["pass"]
+            MAIL_WEB_SERVER = data["web"]
+            logging.info("Using temporary mailing account for email support."
+                         "Watch logs for links with received messages")
+
+    # fail-safe switch in case anything failed above
+    if not MAIL_SERVER:
+        MAIL_SUPPRESS_SEND = True
+        logging.warning("Failed to initialize temporary mailing account! "
+                        "Disabling email support. Emails will not be sent")
+
+
+class ProductionConfig(Config):
+    """
+    Used for https://codeforpoznan.pl/ (AWS Lambda)
+    """
+    pass
+
+
+class StagingConfig(Config):
+    """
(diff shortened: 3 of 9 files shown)


backend/app.py
```diff
@@ -25,7 +27,8 @@
 def create_app():
     """Application factory function"""
     app = Flask(__name__)
-    app.config.from_object("backend.config.DevelopmentConfig")
+    env = os.environ["FLASK_ENV"].title()
```
arturtamborski, line 30: this variable does not exist in aws lambda context yet, we have to add that first to not break the page.

backend/Dockerfile
```diff
@@ -14,7 +14,7 @@ COPY Pipfile Pipfile.lock ./
 
 RUN pip install --upgrade pip \
     && pip install pipenv==v2020.11.15 \
-    && pipenv install --system --deploy
+    && pipenv install --dev --system --deploy --verbose
```
arturtamborski, line 17: I added verbose flag to better understand what's going on, can be removed because the output turned out to not be very useful, but it might come in handy at some point.

backend/config.py
```diff
@@ -1,32 +1,92 @@
...
+    BASE_URL = os.environ["BASE_URL"]
+
     """db config"""
-    DB_USER = os.environ.get("DB_USER")
-    DB_PASSWORD = os.environ.get("DB_PASSWORD")
-    DB_HOST = os.environ.get("DB_HOST")
-    DB_PORT = os.environ.get("DB_PORT") or 5432
-    DB_NAME = os.environ.get("DB_NAME")
+    DB_HOST = os.environ["DB_HOST"]
+    DB_PORT = os.environ["DB_PORT"] or 5432
```
arturtamborski, line 21: this `or 5432` also should be removed but aws lambda does not have `DB_PORT` env variable so I have to fix that first in terraform and then we can remove it :)

backend/config.py
```diff
@@ -1,32 +1,92 @@
...
+
+    """mail config"""
+    MAIL_SUPPRESS_SEND = False  # always off, never block mail sending
+    MAIL_DEBUG = True  # always on, needed for error checking
+    MAIL_USE_TLS = True  # always on, for security (TM)
+    MAIL_USE_SSL = False  # always off, for security (TM)
+    MAIL_WEB_SERVER = None  # defaults to empty, might be set later
+    MAIL_RECIPIENT = "hello@codeforpoznan.pl"
+
+    MAIL_SERVER = os.environ.get("MAIL_SERVER")
```
arturtamborski, line 41: mail config is not critical (as opposed to db config), that's why I'm using `get()`. I assumed that if the config is empty (for example on localhost) then the app still should boot but with additional information in logs. Lambda environments will always have these env variables set, so this config is needed only for the local development.

backend/config.py
```diff
@@ -1,32 +1,92 @@
...
+            "Disabling email support. Emails will not be sent"
+        )
+
+
+class ProductionConfig(Config):
+    """
+    Used for https://codeforpoznan.pl/ (AWS Lambda)
+    """
+
+    pass
```
arturtamborski, line 76: It's important to keep the configs as similar as possible - in this case it's perfect :) Later on when the prod environment will be introduced we might consider flipping few levers here. but for now it should stay the same as any other config.

backend/resources/contact.py
```diff
@@ -1,34 +1,77 @@
...
-            recipients=[f"hello@{base_url}"],
+
+        name, email = data["name"], data["email"]
+        phone, content = data["phone"], data["content"]
+
+        subject = f"""
+            Email z {BASE_URL} od {name}
+        """
+
+        body = f"""
```
arturtamborski, line 37: bit easier to write that way, we can also invest some time in development of HTML templated emails, but this should be done in separate ticket (we don't have templating library installed and it would be out of scope for this one).

backend/resources/contact.py
```diff
@@ -1,34 +1,77 @@
...
+
+    @staticmethod
+    def _send_message(**kwargs):
+        with wrap_io() as (out, err):
+            mail.send_message(**kwargs)
+
+        # known mail testing domain, find message link in output
+        if MAIL_WEB_SERVER:
+            if "MSGID=" in err():
+                msgid = err().split("MSGID=")[1].split("]")[0]
```
arturtamborski, line 65: ethereal mail sends unique message id in SMTP response which we want to parse and print, because it's the key to the message link (see commands below).

backend/resources/contact.py
```diff
@@ -1,34 +1,77 @@
...
+                link = f"https://ethereal.email/message/{msgid}"
+                logging.info(f"Mail sent successfully: {link}")
+                return HTTPStatus.OK
+            else:
+                logging.error("Failed to send mail, SMTP stderr:\n", err())
+                return HTTPStatus.SERVICE_UNAVAILABLE
+
+        # real smtp server
+        else:
+            # TODO: parse SES SMTP server responses like we already do above
```
arturtamborski, line 75: this has to wait for infrastructure ticket

backend/config.py
```diff
@@ -1,32 +1,92 @@
...
+
+    MAIL_SERVER = os.environ.get("MAIL_SERVER")
+    MAIL_PORT = os.environ.get("MAIL_PORT")
+    MAIL_USERNAME = os.environ.get("MAIL_USERNAME")
+    MAIL_PASSWORD = os.environ.get("MAIL_PASSWORD")
+
+    # empty MAIL_* vars are a hint that we should use the temporary account
+    if not MAIL_SERVER:
+        json = {"requestor": "nodemailer", "version": "6.5.0"}
+        resp = requests.post("https://api.nodemailer.com/user", json=json)
```
arturtamborski, line 49: every start of application (in docker) will ask for new temporary user and use that user for sending mails :)
This will let every one of us test the mailing system locally (for example when someone would work on HTML templates) without the need of deploying the code over and over again to staging. Messages are automatically deleted after 12 hours.

arturtamborski:
ToDo: merge this and use it for templating the second email: https://github.com/CodeForPoznan/Community/pull/79/files

arturtamborski:
AWS approved our request to switch SES to production (required if you want to send outgoing e-mails) which means that we're now _almost_ done :)

Few important things to consider:

Acceptable Use Policy (it's short, worth reading)
https://aws.amazon.com/aup/

Service Terms (search for "Simple Email Service", you should find it under number 15)
https://aws.amazon.com/service-terms/

Best Practices:
https://docs.aws.amazon.com/ses/latest/DeveloperGuide/tips-and-best-practices.html

---

One notable bit of information is that we should _not_ use `hello@` for our e-mail address, because `hello@` is a real inbox, so it shouldn't be treated as an automatic one (by SES).

backend/config.py
arturtamborski: Well, almost. I don't like the environment split that we have between three (!!!) different places -there's live-like (lambdas), CI/CD (travis) and local (docker). This at best should be one, but can't be realistically done due to differences between our working environments. Docker in lambda is a bad idea so the last thing to fix is to work on running docker in travis in order to merge these two environments together. Not really a part of this ticket, but should be done in some foreseeable future. Bonus points for switching to github actions :)

backend/config.py
arturtamborski: I'll try to cache that somehow because it's not cool to spam with requests for new accounts constantly during development (flask reload causes it to fetch new user), but if that won't work then I'll have to leave it as is, or maybe turn off the mail support by default.

backend/config.py
arturtamborski: not true anymore - every environment variable ever used in this code _has to be_ clearly defined, there are no defaults. If required env var does not exist, the app will not start.

backend/resources/contact.py
```diff
@@ -1,34 +1,116 @@
...
+        Treść:
+        {content}
+    """
+
+    external_message_subject = """
+        [CfP] Witaj!
+    """
+
+    external_message_body = """
+        Cześć,
```
arturtamborski, line 40: I need someone to send me better idea for this text :)

backend/tests/conftest.py
```diff
@@ -135,10 +135,10 @@ def open(self, *args, **kwargs):
 @pytest.fixture
 def new_msg():
     msg = {
-        "name": "TestName",
-        "email": "test@test.com",
```
arturtamborski, line 139: test.com is a real domain, we don't want to send any mail there :) `example.com` is recommended afaik.

backend/config.py
arturtamborski: Caching wasn't that easy, because flask doesn't provide any interface for that and keeping a local state (file) is too much of a hassle for this problem I think, so instead I disabled mail sending by default so this won't be a problem unless you want to test the specific mail sending thingy.

backend/config.py
arturtamborski: I was thinking that we might even remove the config split completely - it's exactly the same anyway and the only difference between particular environments is in the runtime (lambda vs docker) and in specific env variables, which boils down to secret values only. Just a thought, not really important as the configs _are_ the same anyway.

backend/resources/contact.py
```diff
@@ -1,34 +1,116 @@
...
+
+        except MessageError:
+            return (
+                {"message": "Failed to send contact message"},
+                HTTPStatus.SERVICE_UNAVAILABLE,
+            )
 
         return {"message": "Contact message successfully sent"}, HTTPStatus.OK
+
+    def send_message(self, *, context: dict, subject: str, body: str, **kwargs):
```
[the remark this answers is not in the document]
arturtamborski: forces python to accept keyword only arguments.

```python3
#           first variable   second     third
send_message({'context': 1}, "subject", "body")

the above won't work, you have to call it like so:

send_message(context={'context': 1}, subject="subject", body="body")
```

It's useful for enforcing on api user to call the function with explicit arguments because a person who developed that function might foresee that call to the function might not be obvious. See https://github.com/CodeForPoznan/codeforpoznan.pl_v3/pull/375/files/b4f4c26a90c466b4ceb06573af999efb48ecdc2c#diff-ed7a4a56431b7cc5a9fc51246267b6a34500169fb29baf17ca8883351155e6b7R68 for details - if it would be called without named arguments then it'd be very difficult to understand what's going on.

TLDR: it's a syntax / pretty code thing.

w1stler (approved):
IMO looks okay. Text message can be changed later in separate PR.

Issue #365, opened 2021-03-17
Menu should stay at the top while scrolling


OtisRed:
#### Subject of the issue:
Menu should stick to the top of the display

PR #376, opened 2021-04-16
Fixed header on HomePage

Marce1ina:
#### Story / Bug id:
#365

#### Description:
Added layout markups for vuetify to make the HomePage Header stick to the top of display while scrolling through the page.
Added scrollactive lib to handle changing nav tab to active when scrolling.

#### Migrations:
N/A

https://www.npmjs.com/package/vue-scrollactive

Manual - accordingly to description.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,8 +1,13112 @@
 {
   "name": "frontend",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "0.1.0",
+      "dependencies": {
+        "@mdi/font": "^4.5.95",
+        "axios": "^0.21.1",
+        "vue": "^2.6.10",
+        "vue-router": "^3.3.4",
+        "vue-scrollactive": "^0.9.3",
+        "vuelidate": "^0.7.4",
+        "vuetify": "^2.3.0",
+        "vuex": "^3.0.1"
+      },
+      "devDependencies": {
+        "@vue/cli-plugin-babel": "^3.8.0",
+        "@vue/cli-plugin-eslint": "^3.8.0",
+        "@vue/cli-service": "^3.8.0",
+        "babel-eslint": "^10.0.1",
+        "eslint": "^5.8.0",
+        "eslint-config-prettier": "^4.3.0",
+        "eslint-plugin-import": "^2.17.3",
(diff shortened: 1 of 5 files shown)


KSodowska (approved):
Hi, it runs smoothly :)

KSodowska:
@OtisRed one thing considering header topic: right now on our main page, it highlights your position while scrolling, on localhost - it stays on 'O nas'.

header_main


obraz

Marce1ina:
@KSodowska @OtisRed Scrolling issue should be fixed now, please check again.

KSodowska (approved):
Hi @Marce1ina, navigation tab works :)

KSodowska:
@Marce1ina But at the same time, something minor came up, 'Strona społeczności' title doesn't adjust in a certain conditions (in the master it works fine). Pls check if you have the same 'problem'.

Both are in the same conditions: 
Webrowser: Firefox 88.0, Chrome 90.0 (Ubuntu)
Screen: 1366x768, 1360x768
Zoom: 100% (works fine when it's 90% and less,110% or more)

Master:
Strona społ master

Branch #376
Strona społ branch test_376

Marce1ina:
@KSodowska I'm not able to reproduce this issue on MacOS (tried both Chrome and Firefox in same versions). Does it stay like this after page refresh for you? Think we would need another Ubuntu user to verify this.

OtisRed:
@KSodowska please create a new ticket for this problem as it's out of the scope of this PR, yet still it would be good to investigate.
@Marce1ina it's unblocked now - sorry for the mistake

KSodowska:
@OtisRed hi, I'll do it tomorrow. I want to check it out on RWD simulator first.

stanislawK (approved):
It would be nice to have same behavior without  external package

OtisRed:
@stanislawK thanks for CR! This was a work around because Vuetify couldn't cut it for some reason. Also I think we could use these scroll-events for other purposes. I will create ticket so we can talk about possiblities of solvng this differently but for now I will merge it as a working solution :) 

@Marce1ina you just need to resolve conflivts because they block merging :)

Marce1ina:
@stanislawK I didn't find any build-in solution for this in Vuetify and native intersection-observer API wasn't willing to cooperate so I used the external library. I know it would be nice to not have it, but considering how much time I've already spent on trying to make it work otherwise, it's just not worth the effort.

@OtisRed I've resolved conflicts so if you're ok with that, let's merge this PR.

Issue #309, opened 2020-09-16
Create new participant form - incorrect validation 


jacekkalbarczyk:
**Subject of the issue:**

Validators for Create new participant form are too strict.

**Actual behaviour:**

First name, Last name - polish letters are not allowed.
Slack nick - Underscore  is not allowed.

**Expected behaviour:**

First name, Last name - polish letters should be allowed.
Slack nick - all characters allowed by Slack should be allowed.

**Technical description**

It should be enough to delete certain requirements in vuelidate.
See documentation: https://vuelidate.js.org/#getting-started

PR #377, opened 2021-04-24
Change validation rules in new participant form

Marce1ina:
#### Story / Bug id:
#309

#### Description:
In new participant form:
- removed characters type validation for "First name" and "Last name",
- created custom validators for "GitHub username" and "Slack nick".

#### Migrations:
N/A

N/A

Manual check of front-end form validation as follows:
- "First name" should accept any character,
- "Last name" should accept any charater,
- "Github username" should accept latin alphabet letters, as well as numbers and `-`, with the constraint that username cannot start or end with `-`,
- "Slack nick" should accept latin alphabet letters, as well as polish diacritical characters `żźćńółęąś`, numbers and `,.';-_/()[]{}`.

### DIFF
--- a/frontend/src/components/Participants.vue
+++ b/frontend/src/components/Participants.vue
@@ -183,11 +183,14 @@ import {
   minLength,
   maxLength,
   email,
-  numeric,
-  alpha,
-  alphaNum
+  numeric
 } from 'vuelidate/lib/validators';
 import _ from 'lodash';
+import {
+  slackNickValidator,
+  gitHubUsernameValidator
+} from '../helpers/validation';
+
 export default {
   data() {
     return {
@@ -223,7 +226,7 @@ export default {
       },
       github: {
         required,
-        alphaNum,
+        gitHubUsernameValidator,
         githubExists(value) {
           return this.editParticipant
             ? true
@@ -234,18 +237,16 @@ export default {
       },
       first_name: {
         required,
-        alpha,
         minLength: minLength(3),
         maxLength: maxLength(25)
       },
       last_name: {
         required,
-        alpha,
         minLength: minLength(3),
         maxLength: maxLength(25)
       },
       slack: {
-        alphaNum,
+        slackNickValidator,
         minLength: minLength(3),
         maxLength: maxLength(25)
       },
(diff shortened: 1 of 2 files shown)


frontend/src/components/Participants.vue
```diff
@@ -183,11 +183,14 @@ import {
...
-  numeric,
-  alpha,
-  alphaNum
+  numeric
 } from 'vuelidate/lib/validators';
 import _ from 'lodash';
+import {
+  slackNickValidator,
+  gitHubUsernameValidator
+} from '../helpers/validation';
```
OtisRed, line 192: That's a neat trick to get around it - kudos!

frontend/src/helpers/validation.js
```diff
@@ -0,0 +1,11 @@
+import { helpers } from 'vuelidate/lib/validators';
+
+export const slackNickValidator = helpers.regex(
+  'slackNickValidator',
+  /^[0-9A-Za-zżźćńółęąśŻŹĆĄŚĘŁÓŃ,.';\-_/()[\]{}]*$/
```
OtisRed, line 5: I hate to be THAT guy but I would use: `ąćęłńóśźżĄĆĘŁŃÓŚŹŻ` - just for the sake of my OCD (and readability).

OtisRed (approved):
I pulled the brunch, tried to spoil it, but it seems to work well. From my point of view it solves all the problems mentioned in the task so I leave approve but I would love for sb (@w1stler, @stanislawK, @jacekkalbarczyk, @kristhina) more fluent with regex to check the two experssions in custom validators at the bottom - before we merge it.

frontend/src/helpers/validation.js
Marce1ina: Done

frontend/src/helpers/validation.js
```diff
@@ -0,0 +1,11 @@
+import { helpers } from 'vuelidate/lib/validators';
+
+export const slackNickValidator = helpers.regex(
+  'slackNickValidator',
+  /^[0-9A-Za-ząćęłńóśźżĄĆĘŁŃÓŚŹŻ,.';\-_/()[\]{}]*$/
+);
+
+export const gitHubUsernameValidator = helpers.regex(
+  'gitHubUsernameValidator',
+  /^[0-9A-Za-z][0-9A-Za-z-]*[0-9A-Za-z]$/
```
stanislawK, line 10: Whenever You want match the character - literally with regex, it is safer to escape it and use "\-" instead of "-".

Marce1ina: @stanislawK I have removed it as it causes lint failure: `error: Unnecessary escape character: \- (no-useless-escape) at src/helpers/validation.js:10:26`

Issue #371, opened 2021-04-14
Participants list and dropdown alphabetical order


stanislawK:
#### Subject of the issue:
For HacknightParticipants component both dropdown and list of participants should have alphabetical order.

#### Your environment:
* version/branch
* which browser and its version

#### Steps to reproduce:
Describe how to reproduce this issue

#### Actual behaviour:
Describe what happened

#### Expected behaviour:
For HacknightParticipants component both dropdown and list of participants should have alphabetical order.

PR #384, opened 2021-05-05
[371] Participants alphabetical order (fixes#371)

stanislawK:
#### Story / Bug id:
[371](https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/371)

#### Description:
Participants should be now order alphabetically by github in both hacknight's participants list and participant dropdown.

#### Migrations:
N/A

N/A

From dashboard component select hacknight, check whether participants list is in alphabetical order, try to add new participant - github names from dropdown should be sorted too.

### DIFF
--- a/frontend/src/components/HacknightsParticipants.vue
+++ b/frontend/src/components/HacknightsParticipants.vue
@@ -5,16 +5,7 @@
         <v-combobox
           v-model="selectedParticipants"
           outlined
-          :items="
-            getParticipants
-              .filter(
-                participant =>
-                  !getHacknight.participants.filter(
-                    y => y.id === participant.id
-                  ).length
-              )
-              .sort((a, b) => a.github.localeCompare(b.github))
-          "
+          :items="filerOutParticipants()"
           item-text="github"
           item-value="github"
           :search-input.sync="search"
@@ -94,6 +85,26 @@ export default {
         .dispatch('hacknight/addParticipants', ids)
         .then(() => (this.selectedParticipants = []));
     },
+    // filerOutParticipants() {
+    //     return this.getParticipants
+    //       .filter(
+    //         participant =>
+    //           !this.getHacknight.participants.filter(
+    //             y => y.id === participant.id
+    //           ).length
+    //       )
+    //       .sort((a, b) => a.github.localeCompare(b.github))
+    //   },
+      filerOutParticipants() {
+        return this.getParticipants
+          .filter(
+            participant =>
+              !this.getHacknight.participants.some(
+                hacknightParticipant => hacknightParticipant.id === participant.id
+              )
+          )
+          .sort((a, b) => a.github.localeCompare(b.github))
+      },
     orderedParticipants() {
       return _.sortBy(this.getHacknight.participants, participant =>
         participant.github.toLowerCase()
--- a/frontend/src/components/HacknightsParticipants.vue
+++ b/frontend/src/components/HacknightsParticipants.vue
@@ -6,11 +6,14 @@
           v-model="selectedParticipants"
           outlined
           :items="
-            getParticipants.filter(
-              participant =>
-                !getHacknight.participants.filter(y => y.id === participant.id)
-                  .length
-            )
+            getParticipants
+              .filter(
+                participant =>
+                  !getHacknight.participants.filter(
+                    y => y.id === participant.id
+                  ).length
+              )
+              .sort((a, b) => a.github.localeCompare(b.github))
           "
           item-text="github"
           item-value="github"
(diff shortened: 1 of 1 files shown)


KSodowska (approved):
Hi, it works everywhere -  on Add participant to a hacknight drop down, on a Participant list and on Select participant drop down on Edit existing participant :)

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -59,7 +62,7 @@
       </v-toolbar>
       <v-divider v-if="getHacknight.participants"></v-divider>
       <v-list>
-        <template v-for="(item, i) in getHacknight.participants">
+        <template v-for="(item, i) in orderedParticipants()">
```
Marce1ina, line 65: Not a big deal in this case, but it would probably be more accurate to put `orderedParticipants` in `computed` instead of `methods`.

> Instead of a computed property, we can define the same function as a method. For the end result, the two approaches are indeed exactly the same. However, the difference is that computed properties are cached based on their reactive dependencies. A computed property will only re-evaluate when some of its reactive dependencies have changed.

From https://vuejs.org/v2/guide/computed.html

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -6,11 +6,14 @@
...
-                  .length
-            )
+            getParticipants
+              .filter(
+                participant =>
+                  !getHacknight.participants.filter(
+                    y => y.id === participant.id
+                  ).length
+              )
+              .sort((a, b) => a.github.localeCompare(b.github))
```
Marce1ina, line 16: I know it wasn't perfect before but after adding another line this whole thing gets really unreadable and it clutters the template. Please move this logic to `computed`, perhaps as `participantSelectItems`.

It would also be cool to refactor filter function a bit, to make it more readable. Think this would do the work:

```js
participant => !getHacknight.participants.some(
    hacknightParticipant => hacknightParticipant.id === participant.id
)
```

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -89,6 +93,11 @@ export default {
       this.$store
         .dispatch('hacknight/addParticipants', ids)
         .then(() => (this.selectedParticipants = []));
+    },
+    orderedParticipants() {
+      return _.sortBy(this.getHacknight.participants, participant =>
```
Marce1ina, line 98: Any reason why `localeCompare` wouldn't handle this? Seems like we could even have one helper function defined in `methods`, that would handle sorting and then use it for both places where it's needed in this component. That would be more DRY solution.

Marce1ina (changes requested):

frontend/src/components/HacknightsParticipants.vue
stanislawK: I have tried that and it didn't work (offcourse I could mess somethnig up). Since we are recieving participants using getter (getHacknight) within computed I have had no access to participants from another method computed.

Marce1ina: I think it should be possible to use one computed inside another. Would `this` help? https://stackoverflow.com/questions/41173998/is-it-possible-to-use-the-computed-properties-to-compute-another-properties-in-v

OtisRed:
@stanislawK I guess there are still few comments waiting for you here :) Just a reminder!

frontend/src/components/HacknightsParticipants.vue
stanislawK: I would not - `this` is already there.  With that approach we are attempting to order participants which we have not got yet.

frontend/src/components/HacknightsParticipants.vue
stanislawK: In my opinion sortBy is more descriptive and easier to follow, it is also more stable sorting algorithm and in most cases faster. In one method we want to order participants, in another filter them out, so for me at least, these are different functionalities, and I would prefer to keep them separately.

frontend/src/components/HacknightsParticipants.vue
Marce1ina: I've checked out your branch, moved `orderedParticipants` to `computed` and it works fine if you wrap function body into `if(this.getHacknight) {...}`.

frontend/src/components/HacknightsParticipants.vue
Marce1ina: I'm just asking because you've used `localeCompare` above. In current version it's line 96. Methods you're talking about are different indeed but they have the common sorting part, which I'm suggesting to extract.

Marce1ina:
Let me know if you'd like to discuss my comments in real time, should be much easier for both of us :) I could also propose a commit with changes I have in mind so you could decide if they look fine to you, don't want to waste our time on prolonged written discussions.

Marce1ina (approved):

Issue #56, opened 2019-05-08
Create search bar for participant component


OtisRed:
#### General description
We should be able to search participants by any kind of data to find other details when needed.

#### Technical description
Ideally, a user should be able to type any kind of data (name, surname, github, slack) in the same field and get the probable result.  Given that this function is relevant both for editing details in `participants` component and displaying details in #103 we could extract it in the separate re-usable component. 

#### Resources
We could use the combobox solution like [this](https://github.com/CodeForPoznan/codeforpoznan.pl_v3/blob/master/frontend/src/components/HacknightsParticipants.vue) but without multiple objects as developed 

#### To-do list
- [ ] Extract search function in separate component
- [ ] Use it in participants component (create/edit form has it at the top) and replace v-select to unify the use
- [ ] Make note to use the same solution in #103

OtisRed:
Waiting for the proposal for general design of the backoffice

OtisRed:
@Marce1ina would you like to take this one next? It's a bit more challanging I guess but it also makes more sense to develop #103 after having this one done

Marce1ina:
@OtisRed I can take this one next.

PR #410, opened 2021-06-02
Create ParticipantsSearch component and use it in Participants form

Marce1ina:
#### Story / Bug id:
https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/56

#### Description:
Added ParticipantsSearch component and used it in Participants form to replace v-select.

#### Migrations:
N/A

N/A

1. Go to 'Edit existing participant' form and make sure that participant can be found in searchbox by any related data (slack, github, name, last name, email, phone number, id). 
2. Sanity testing of all the functionalities in this form (ie. you can edit participant, save changes, then search by changed data; you can clear the form; you can switch between 'edit' and 'add' mode without any disturbance).

### DIFF
--- a/frontend/src/components/ParticipantsSearch.vue
+++ b/frontend/src/components/ParticipantsSearch.vue
@@ -1,0 +1,53 @@
+<template>
+  <v-combobox
+    clearable
+    label="Search participants"
+    :items="sortedParticipants"
+    item-text="github"
+    :filter="filterParticipants"
+    v-model="selectedParticipant"
+  ></v-combobox>
+</template>
+
+<script>
+import { mapGetters } from 'vuex';
+
+export default {
+  props: {
+    value: {
+      type: Object | null,
+      required: true
+    }
+  },
+  computed: {
+    ...mapGetters('participant', {
+      getParticipants: 'getParticipants'
+    }),
+    sortedParticipants() {
+      return this.getParticipants.sort((a, b) =>
+        a.github.localeCompare(b.github)
+      );
+    },
+    selectedParticipant: {
+      get() {
+        return this.value;
+      },
+      set(selectedParticipant) {
+        this.$emit('input', selectedParticipant);
+      }
+    }
+  },
+  methods: {
+    filterParticipants(item, queryText) {
+      return Object.values(item).some(
+        value =>
+          value &&
+          value
+            .toString()
+            .toLowerCase()
+            .includes(queryText.toLowerCase())
+      );
+    }
+  }
+};
+</script>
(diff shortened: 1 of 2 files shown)


KSodowska (approved):
@Marce1ina lookin' good :+1:

frontend/src/components/ParticipantsSearch.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+  <v-combobox
+    clearable
+    label="Search participants"
+    :items="sortedParticipants"
+    item-text="github"
```
OtisRed, line 6: I know that you replicated the previous solution here but it would make user's life much easier if we could render/display results for each item in the format:
"`first_name` `last_name` (`slack`/`github`)"

In the documentation I found that combobox accepts stings:
https://vuetifyjs.com/en/api/v-combobox/#props
Not sure if that applies to the example above. 

Yet, given that we will use this component for different purposes and there are three different sources of information about who selectedParticipant is it would be great to have all that information on the list not to wonder about if the person with slack nick A is also the person with github nick B (and they do vary)

frontend/src/components/ParticipantsSearch.vue
```diff
@@ -0,0 +1,52 @@
...
+    }
+  },
+  computed: {
+    ...mapGetters('participant', {
+      getParticipants: 'getParticipants'
+    }),
+    sortedParticipants() {
+      const participants = this.getParticipants;
+
+      return participants.sort((a, b) => a.github.localeCompare(b.github));
```
OtisRed, line 28: If the second would be feasible we could use `first_name` to sort participants instead of github. 
(Knowing someone by name comes naturally and if the pattern I described before is possible, then this additional information would make it easy to distinguish between different people having the same name)

OtisRed (changes requested):
One technical comments and two regarding user experience :)

Marce1ina:
@OtisRed thanks for your input, I've addressed the comments, please check if you're fine with the new behaviour.

OtisRed:
@Marce1ina thanks! It's much easier to navigate now! However two more points before we merge it:
1. I guess it's important to figure out the consistent solution between this and #384 so if possible please talk about it with @stanislawK.
2. Would it be possible to show "no results" in case when the input has no match? Having information that there is no information can be also important information :D Also I think this might be related to the discussion you have with @stanislawK about filtering and sorting functions.

Marce1ina:
@OtisRed added 'no results' message. The consistency issue will be addressed separately, as discussed on priv.
Will need your approval to merge :)

OtisRed (approved):
Perfect!

Issue #364, opened 2021-03-17
Delete participant from hacknight 


OtisRed:
#### Subject of the issue:
In the list of participants for particular hacknight there is no option to delete them from it. We should add an option to delete participants. E.g. we could create "X" button for each participant.

PR #417, opened 2021-06-20
Added delete buttons for hacknight participants

jacekkalbarczyk:
#### Story / Bug id:
#364 

#### Description:
Added buttons enabling removing participants from hacknight at HacknightsParticipants component. 

#### Migrations:
N/A

N/A

N/A

### DIFF
--- a/backend/resources/hacknight.py
+++ b/backend/resources/hacknight.py
@@ -91,13 +91,13 @@ class HacknightParticipants(Resource):
         hacknight = Hacknight.query.get_or_404(id)
         participants = {participant.id for participant in hacknight.participants}
         json_data = request.get_json(force=True)
-        ids_schema = Schema.from_dict({"participants_ids": fields.List(fields.Int())})
+        ids_schema = Schema.from_dict({"participant_id": fields.List(fields.Int())})
         try:
             data = ids_schema().load(json_data)
         except ValidationError as err:
             return err.messages, HTTPStatus.UNPROCESSABLE_ENTITY
 
-        to_remove = participants.intersection(set(data["participants_ids"]))
+        to_remove = participants.intersection(set(data["participant_id"]))
         if not to_remove:
             return {"message": "No participant to delete"}, HTTPStatus.BAD_REQUEST
         hacknight.participants = [
--- a/backend/resources/hacknight.py
+++ b/backend/resources/hacknight.py
@@ -91,13 +91,13 @@ class HacknightParticipants(Resource):
         hacknight = Hacknight.query.get_or_404(id)
         participants = {participant.id for participant in hacknight.participants}
         json_data = request.get_json(force=True)
-        ids_schema = Schema.from_dict({"participant_id": fields.List(fields.Int())})
+        ids_schema = Schema.from_dict({"participant_ids": fields.List(fields.Int())})
         try:
             data = ids_schema().load(json_data)
         except ValidationError as err:
             return err.messages, HTTPStatus.UNPROCESSABLE_ENTITY
 
-        to_remove = participants.intersection(set(data["participant_id"]))
+        to_remove = participants.intersection(set(data["participant_ids"]))
         if not to_remove:
             return {"message": "No participant to delete"}, HTTPStatus.BAD_REQUEST
         hacknight.participants = [
--- a/backend/resources/hacknight.py
+++ b/backend/resources/hacknight.py
@@ -91,13 +91,13 @@ class HacknightParticipants(Resource):
         hacknight = Hacknight.query.get_or_404(id)
         participants = {participant.id for participant in hacknight.participants}
         json_data = request.get_json(force=True)
-        ids_schema = Schema.from_dict({"participant_ids": fields.List(fields.Int())})
+        ids_schema = Schema.from_dict({"participants_ids": fields.List(fields.Int())})
         try:
             data = ids_schema().load(json_data)
         except ValidationError as err:
             return err.messages, HTTPStatus.UNPROCESSABLE_ENTITY
 
-        to_remove = participants.intersection(set(data["participant_ids"]))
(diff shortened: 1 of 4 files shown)


w1stler (approved):
Seems okay. Do we have any tests for that?

jacekkalbarczyk:
We don't.

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -65,6 +65,7 @@
               <v-icon>mdi-account-outline</v-icon>
             </v-list-item-avatar>
             <v-list-item-title v-text="item.github"></v-list-item-title>
+            <v-btn v-on:click="onDeleteParticipant(item)">Delete</v-btn>
```
OtisRed, line 68: Could we consider sth more graphically aesthetic than DELETE? :D We have [font awsome oboard](https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=delete) (e.g. see the exit button in `ModalContent.vue`) so we could use sth like these:
* https://fontawesome.com/v5.15/icons/user-times?style=solid
* https://fontawesome.com/v5.15/icons/backspace?style=solid

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -89,6 +90,11 @@ export default {
       this.$store
         .dispatch('hacknight/addParticipants', ids)
         .then(() => (this.selectedParticipants = []));
+    },
+    onDeleteParticipant(participant) {
+      const ids = [participant.id];
```
OtisRed, line 95: You probably mirrored the solution for adding the participants. However @stanislawK used there `ids` in plural because there are many participants that can be added at once, but only one that can be deleted at a time given this solution we agreed upon. So please use sth like` participant_id` in singular form to name the variable and keep it consistent as you pass it to Vuex so that no one has to wonder if this is the same thing under the different names :D

frontend/src/store/modules/hacknight.js
```diff
@@ -76,6 +76,20 @@ export default {
         .catch(error => {
           commit('raiseError', error);
         });
+    },
+    deleteParticipants({ commit, getters }, participants_ids) {
```
OtisRed, line 80: ditto => `deleteParticipant` in singular

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -89,6 +90,11 @@ export default {
...
+
+      this.$store.dispatch('hacknight/deleteParticipants', ids);
```
OtisRed, line 97: Ditto => `participant_id`

frontend/src/store/modules/hacknight.js
```diff
@@ -76,6 +76,20 @@ export default {
...
+      axios
+        .delete(`/api/hacknights/${getters.getHacknight.id}/participants/`, {
+          data: {
+            participants_ids: participants_ids
```
OtisRed, line 84: ditto => `participant_id: participant_id`

OtisRed (changes requested):
It works exactly how it should - I tested PR locally, tried to break it in couple of ways and didn't accomplish shit - which is good. Nevertheless, I leave one aesthetic and some semantic comments for the sake of design and future development :D

frontend/src/components/HacknightsParticipants.vue
KSodowska: > Could we consider sth more graphically aesthetic than DELETE? :D We have [font awsome oboard](https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=delete) (e.g. see the exit button in `ModalContent.vue`) so we could use sth like these:
> 
>     * https://fontawesome.com/v5.15/icons/user-times?style=solid
> 
>     * https://fontawesome.com/v5.15/icons/backspace?style=solid

@OtisRed agreed - these DELETE button is kind of a psycho :D

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -65,6 +65,9 @@
               <v-icon>mdi-account-outline</v-icon>
             </v-list-item-avatar>
             <v-list-item-title v-text="item.github"></v-list-item-title>
+            <v-btn v-on:click="onDeleteParticipant(item)">
```
OtisRed, line 68: How would you like this?
```suggestion
            <v-btn icon v-on:click="onDeleteParticipant(item)">
```

frontend/src/components/HacknightsParticipants.vue
```diff
@@ -65,6 +65,9 @@
...
+              <i class="fas fa-user-times fa-lg"></i>
```
OtisRed, line 69: Maybe we could also use red color to communicate deleteing function?
```suggestion
              <i class="button_delete fas fa-user-times fa-lg"></i>
```
Then we could add new class:
```
.button_delete {
  color: $red;
}
```
And add global red (e.g. #ed130c) in `scss/colors.scss` so it can parse it from a global source :)

OtisRed (approved):
Looks good to me! I left two comments regarding icon and color but I can take it how it is. I leave approve so you can decide and merge it afterwards :)

OtisRed:
Also - you got conflicts :P

backend/resources/hacknight.py
```diff
@@ -91,13 +91,13 @@ def delete(self, id):
         hacknight = Hacknight.query.get_or_404(id)
         participants = {participant.id for participant in hacknight.participants}
         json_data = request.get_json(force=True)
-        ids_schema = Schema.from_dict({"participants_ids": fields.List(fields.Int())})
+        ids_schema = Schema.from_dict({"participant_id": fields.List(fields.Int())})
```
stanislawK, line 94: We should change participants_ids -> participant_id in post method too. Same for all unit tests in which we have "participants_ids". After those changes all tests should pass.

KSodowska (approved):

Issue #355, opened 2021-02-03
Create a chart showing the number of participants in time


OtisRed:
#### Subject of the issue:
Since we have database working we could employ that to visualize its content to provide information that is easy for the eye and possible to compare. I would suggest to start with a simple line chart that show number of participants in time (e.g. three months by default). 

I did some googling and this have caught my eye:
https://apexcharts.com/docs/chart-types/line-chart/ 

But also here is a good piece about other libraries that are easy to use in Vue:
https://medium.com/dailyjs/data-visualization-libraries-for-vue-js-in-2020-c74fb83c5778

PR #423, opened 2021-07-08
Add apexcharts lib and ParticipantsChart

Marce1ina:
#### Story / Bug id:
https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/355

#### Description:
Added libs for data visualization. 
Added Participants in time chart at the bottom of the dashboard page.
Modified response in for _/api/hacknights/_ call to send the participants list.

#### Migrations:
N/A

- vue-apexcharts
- apexcharts

On dashboard, at bottom, there should be a new chart showing number of participants in time. Validate shown data against API (check hacknights/ call in devTools Network tab and compare with the information in the chart).

Zrzut ekranu 2021-08-18 o 20 57 14
Zrzut ekranu 2021-08-18 o 20 57 27
Zrzut ekranu 2021-08-18 o 20 59 47


OtisRed:
Looks good and works nice locally! 

However is there a way to decide dynamically the range for the dates? I mean is it possible to define X axis according to the user input or pre-defined categories like: "lastMonth", "lastQuarter", "lastSixMonths", "last Year" and "allTheHistory"? 

Or is this beyond the scope of this ticket just to implement the working solution? 

In the second case I would create a new ticket for that but I'd like to know your opinion beforehand :)

KSodowska:
@Marce1ina looks awesome :) but I can't generate chart with some historic data (e.g. hacknights from 2019). There's no way to shift the chart to the left (like slider). When I select hacknight from 2019, chart stays the same. I'd appreciate a guidance :D

Marce1ina:
@OtisRed I think it's doable but there's no easy solution built in apexcharts so it won't be a one-liner change. Also, I won't be available for next two weeks so I think it's better to have separate ticket for this change and close this one to avoid outdating the branch.

@KSodowska This chart is not connected to hacknights component, please check the ticket description, it was only about displaying some fixed data. We put it beneath the hacknights component on dashboard only because there was no better place for it.

backend/resources/hacknight.py
```diff
@@ -15,7 +15,7 @@
 class HacknightList(Resource):
     @jwt_required
     def get(self):
-        hacknight_schema = HacknightSchema(many=True, exclude=("participants",))
+        hacknight_schema = HacknightSchema(many=True)
```
stanislawK, line 18: Most likely it is topic for another ticket, but I am bit concern how big response might become after allowing to parse all participants for all hacknights. We will end up with bunch of repeated data (most likely these same participant object will be nested multiple times in different hacknight objects), which we don't really need - right now we are interested in number of participants only.

OtisRed:
I agree, let's put time-scoping for another ticket. Also @stanislawK made a valid point about data use in this solution so we'll discuss it while you're away. Have a good holidays!

Marce1ina:
@OtisRed In the end I've added time-scoping to this PR as it is still not merged. Are there any conclusions on the API change topic?

OtisRed (approved):
I've checked it some time ago already. I re-run the Travis build to make sure everything is in place. I will merge it but I will also leave the ticket #434 for us to discuss handling of data as suggested in the comment by @stanislawK.

Issue #382, opened 2021-05-05
Enhance semantic html tags


OtisRed:
#### Subject of the issue:
I guess we use mostly `<div>` for anything. Perhaps it would be good to audit html templates in Vue and consider richer semantic tagging for SEO purposes as well as accessibility.

PR #431, opened 2021-07-28
Semantic html [WIP] 

Gasimli:
#### Story / Bug id:
[#382](https://github.com/codeforpoznan/codeforpoznan.pl_v3/issues/382)

#### Description:
I Changed the div tags to Semantic tags

#### Migrations:
N/A

N/A

N/A

### DIFF
--- a/frontend/src/components/Dashboard.vue
+++ b/frontend/src/components/Dashboard.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <header>
     <app-dashboard-header />
     <v-row xs12>
       <v-col>
@@ -10,7 +10,7 @@
         <app-participants />
       </v-col>
     </v-row>
-  </div>
+  </header>
 </template>
 
 <script>
--- a/frontend/src/components/DashboardHeader.vue
+++ b/frontend/src/components/DashboardHeader.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <nav>
     <v-app-bar class="navbar-custom" dark clipped-right color="#2C3E50">
       <v-toolbar-title>
         <a href="https://codeforpoznan.pl/">
@@ -37,7 +37,7 @@
         </v-tabs>
       </v-expand-transition>
     </v-list>
-  </div>
+  </nav>
 </template>
 
 <script>
--- a/frontend/src/components/Header.vue
+++ b/frontend/src/components/Header.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <nav>
     <v-app-bar class="navbar-custom" dark color="#2C3E50">
       <v-toolbar-title>
         <a href="https://codeforpoznan.pl/">
@@ -48,7 +48,7 @@
         </v-tabs>
       </v-expand-transition>
     </v-list>
-  </div>
+  </nav>
 </template>
 
 <script>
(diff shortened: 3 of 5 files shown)


frontend/src/components/Header.vue
```diff
@@ -48,7 +48,7 @@
         </v-tabs>
       </v-expand-transition>
     </v-list>
-  </div>
+  </nav>
```
OtisRed, line 51: Check the comment above - these tags are already handled by Vue

frontend/src/components/SocialMedia.vue
```diff
@@ -8,7 +8,7 @@
         </a>
       </li>
     </ul>
-  </div>
+  </address>
```
OtisRed, line 11: Good catch! Well done :)

frontend/src/components/DashboardHeader.vue
```diff
@@ -37,7 +37,7 @@
         </v-tabs>
       </v-expand-transition>
     </v-list>
-  </div>
+  </nav>
```
OtisRed, line 40: Good intuition! Although this would be the good place to use <header> and then <nav>. BUT! Both these tags are already handled by Vue - you can inspect it using devtools (press F12 on the page to explore its code. Works also on the localhost):

So here we have `<header>` rendered by `v-app-bar`:
Zrzut ekranu z 2021-07-28 19-20-47

And few lines below we have `<nav>` rendered by `v-tabs` already:
Zrzut ekranu z 2021-07-28 19-21-29

OtisRed (commented):
Just a reminder - this is an audit task so it's done when you say it's done (or when you don't want to work on it anymore :P ). If you still working on it just edit the title and add [WIP] which means "Work in progress" :) 

Couple of tips:
* In my comments you can see that some of the semantic structure is already handled by Vue but some is not. If you see that we could use some tag in particular place please check in the devtools (press F12 on the website) and check if the tag you have in mind is actually missing. 
* For inspiration you can check the e-book I dent you and/or https://developer.mozilla.org/en-US/docs/Web/HTML 

Few hints:
* Since we have `<header>` and `<footer>` you may wanna find the place for `<main>` tag
* You may consider introducing `<section>` tag to each component on the website (besides header and footer)
* You may check all the lists use `<ul>` and `<li>` tag (in Vue we use v-for directive to generate lists so you may look for it in the codebase)

frontend/src/components/JoinUs.vue
```diff
@@ -16,7 +16,7 @@
             Kodujesz? Jesteś grafikiem? UX-owcem? Robisz cokolwiek innego, co
             może się przydać w kodowaniu dla Poznania? Dołącz do nas! Spotykamy
             się w każdą środę od godz. 18:00 w firmie STX Next przy ul. Mostowej
-            38 niedaleko Nowej Gazowni (Pawilonu) w parku Stare Koryto Warty.
+            38 niedaleko Nowej Gazowni (Pawilonu) w parku Stare Koryto Warty. TEST
```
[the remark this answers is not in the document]
Gasimli: Sorry about that.

frontend/src/components/Dashboard.vue
```diff
@@ -10,7 +10,7 @@
         <app-participants />
       </v-col>
     </v-row>
-  </div>
+  </header>
```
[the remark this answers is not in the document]
Gasimli: Alright,will do that.

OtisRed:
This was replaced by #439 so I close it

Issue #428, opened 2021-07-14
Refactor e-mail handling resources


OtisRed:
#### Subject of the issue:
After implementing #426 and #427 it would be good to clean up the files related to that function.

#### Technical description: 
* In `backend/config.py` some of the code between 33 and 68 line will become redundant or possible to make more clear.
* In `travis.yml` it would be nice **not to** ask etheral email for credentials on production env (we don't test on production!)
* Some of these changes along with #426 and #427 have implications for Terraform in [infrastructure repo](https://github.com/CodeForPoznan/Infrastructure).

PR #435, opened 2021-08-17
Fix email sending code

arturtamborski:
#### Story / Bug id:
https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/428

#### Description:
Make the damn thing work :)

It relies on the PR in Infra: https://github.com/CodeForPoznan/Infrastructure/pull/87
Anyway, it should work just fine from what I tested :)

#### Migrations:
N/A

N/A

N/A


.envrc.example
```diff
@@ -1,7 +1,21 @@
...
 export DB_HOST=db
 export DB_PORT=5432
 export DB_NAME=cfp_v3
 export DB_USER=cfp_v3
 export DB_PASSWORD=cfp_v3
 
+# mail
+export MAIL_SUPPRESS_SEND=FALSE
+export MAIL_USERNAME=
+export MAIL_PASSWORD=
```
OtisRed, line 19: Should this be empty?

backend/config.py
```diff
@@ -27,44 +27,51 @@ class Config:
...
     JWT_BLACKLIST_ENABLED = True
 
     """mail config"""
-    MAIL_DEBUG = True  # always on, needed for error checking
-    MAIL_USE_TLS = True  # always on, for security (TM)
-    MAIL_USE_SSL = False  # always off, for security (TM)
-    MAIL_WEB_SERVER = ""  # defaults to empty str, might be set later
+    MAIL_DEBUG = True  # always on, for error checking
+    MAIL_USE_TLS = True  # always on, for security
+    MAIL_USE_SSL = False  # always off, for security
```
OtisRed, line 37: Shouldn't SSL be good for security? :D

backend/config.py
```diff
@@ -27,44 +27,51 @@ class Config:
...
+    MAIL_WEB_URL = ""  # used only for ethereal.mail
 
-    MAIL_SERVER = os.environ["MAIL_SERVER"]
-    MAIL_PORT = os.environ["MAIL_PORT"]
+    MAIL_SUPPRESS_SEND = os.environ["MAIL_SUPPRESS_SEND"].lower() == "true"
```
OtisRed, line 40: Do you really translate boolean to string here? :D Not the critique => just curious :)

backend/helpers.py
```diff
@@ -23,18 +24,19 @@ def wrap_io(catch=Exception) -> (callable, callable):
         err.write(traceback.format_exc())
 
 
-def ask_for_temporary_mail():
+def ask_for_temporary_mail() -> Optional[dict]:
+    """
+    Return dict with connection info for ethereal.mail or None if failed.
+    """
```
OtisRed, line 30: If I understand right this is the message in case etheral data does not load. If I'm right - it took me a while to figure it out. If I'm wrong - it took me a while not to figure it out anyway. Would it make sense to be a bit more descriptive here?

backend/resources/contact.py
```diff
@@ -109,11 +112,15 @@ def send_message(self, *, context: dict, subject: str, body: str, **kwargs):
 
         stderr = err()
 
-        if "MSGID=" not in stderr:
+        # example successful responses from ethereal and ses, respectively:
+        # data: (250, b'Accepted [STATUS=new MSGID=YRbuVPOb3gQndRMOYRubW3BtxAZI8iBfAAAAAbdiFY1fwZ4LCwdvaNWoIDs]')
+        # data: (250, b'Ok 0102017b4c5776db-6639a4be-4c46-469b-b528-edf2b28e686a-000000')
```
OtisRed, line 117: I'm not really sure what it is for?

OtisRed (approved):
I was rather curious than making any real comments here :D

.envrc.example
arturtamborski: yes, the only real useful case when these vars should be filled is when the code is executed on production environment because that's when the vars will contain real access credentials to SES. On local env the vars should stay empty so that we could use ethereal dynamic credentials instead.

backend/config.py
arturtamborski: good point, that's probably a mistake? will check that

backend/config.py
arturtamborski: quite the opposite - I translate string (from env var) to boolean. It's a cheap way of doing that but works nicely. Env variables don't support types other than string, so that's why I'm doing it.

backend/helpers.py
arturtamborski: Yes, in case ethereal dies/closes/stops working/api changes/whatever out of order happens - we return None which will indicate us that we failed to fetch credentials and we should turn off mail sending. I'll expand this docstring a bit.

backend/resources/contact.py
arturtamborski: just a reference material, it's easier to take a look at the raw data from commends than set up whole environment (especially the production one) just to check how these responses actually look like inside.

backend/config.py
arturtamborski: nope, probably not a mistake. It's a default from flask_mail and it seems to make sense. Ethereal doesn't work with SSL on (though is happy to use TLS option above). Maybe something with ports? Not tested with SES.
I'm happy to leave it as is unless you want me to check it again.

PR #439, opened 2021-08-29
Semantic html [WIP]

Gasimli:
#### Story / Bug id:
#382

#### Description:
Added section tags

#### Migrations:
N/A

N/A

N/A


OtisRed:
There were problems we discussed in the summer and I don't wanna keep it hanging. If @Gasimli decide to work on it then just create another PR :)

PR #442, opened 2021-09-01
Broadened input field (fixes #370)

### LINKED ISSUE #370: Shorten the participant's name inside v-chip
#### Subject of the issue:
As @KSodowska spotted:

> Adding a participant with a name longer than 12 characters cuts-off ‘x’ button, so it's very hard or impossible to remove a participant.

#### Your environment:
Firefox

#### Steps to reproduce:
Try to add to hacknight participants with long name

#### Actual behaviour:
'x' button is not visible for participants with long name

#### Expected behaviour:
'x' button is visible for each participant name. Either participant name inside v-chip is shorten, or v-chip dynamically adjusts to name length. We could also make the combobox a bit broader, because right noe it's very slim.

OtisRed:
#### Story / Bug id:
[370](https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/370)

#### Description:
I broadened the input - doesn't work so much on very small screens but I would to tweak with vuetify insides to target v-chip and alter it. I think it's an easy fix that works for most cases.

### DIFF
--- a/frontend/src/components/HacknightsParticipants.vue
+++ b/frontend/src/components/HacknightsParticipants.vue
@@ -1,8 +1,9 @@
 <template>
   <v-container fluid text-xs-center>
     <v-row align="center" justify="center">
-      <v-col class="d-flex" cols="12" sm="6">
+      <v-col class="d-flex bar" cols="12" lg="8" md="10">
         <v-combobox
+          class="bar"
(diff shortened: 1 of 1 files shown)


jacekkalbarczyk (approved):
It's like you said, it'll work for most cases.

Issue #385, opened 2021-05-05
Modify script for deployment


OtisRed:
#### Subject of the issue:
As we are getting ready to deploy we need to set up how the changes will be pushed onto production. 

#### To-do list
- [ ] Set up production server - replace v2 with v3
- [ ] Prepare infrastructure to work in release cycles
- [ ] Keep dev. server for testing purposes (all changes in master should be pushed automatically here)

arturtamborski:
heyy that's a lot of PRs : )  I finally finished this issue in latest pr #464

PR #457, opened 2022-03-12
replace dynamic cloudfront id fetching with env var

arturtamborski:
fix a bug in #385 :)


deploy.sh
```diff
@@ -24,26 +26,13 @@ case "${1:-}" in
...
 
 
 echo "refresh CDN"
-dist_id=$(
-  aws cloudfront list-distributions \
-    | jq --arg PATH "/${RESOURCE}" -r '
-      .DistributionList.Items[]
-      | {Id, p: .Origins.Items[].OriginPath}
-      | select(.p == $PATH).Id'
-)
```
arturtamborski, line 45: I'm removing (or rather, reverting) this bit because IAM user we're using does not have the permission to list cloudfront distros, and we don't want to give that permission to it, because it's not worth it - instead we'll use another env var and type the distro id directly in there.

PR #458, opened 2022-03-12
simplify installation of dependencies

arturtamborski:
#385


deploy.sh
```diff
@@ -47,7 +47,8 @@ aws cloudfront create-invalidation --paths "/*" --distribution-id "${dist_id}"
 
 
 echo "bundle application"
-(cd backend  && pipenv run pip install -r <(pipenv lock -r) --target ../packages)
```
arturtamborski, line 50: `<(pipenv lock -r)` sometimes fails, so I'm changing it back to regular file write instead.

PR #495, opened 2022-11-09
Fix deployment - jinija2 & markupsafe

w1stler:
#### Story / Bug id:
No ticket

#### Description:
Fixed deployment. We would need to bump other packages as well, but this is to unblock other work.

#### Migrations:
N/A

N/A

N/A


arturtamborski (approved):

OtisRed (approved):
Looked through it. LGTM

Issue #506, opened 2023-01-11
Add simple backend linter to PR


w1stler:
Black & isort on PR

PR #507, opened 2023-01-11
[CFP-506] Add pr backend linter

w1stler:
#### Story / Bug id:
[123](https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/506)

#### Description:
Added PR linter for backend

### DIFF
--- a/.github/workflows/pr_linting.yml
+++ b/.github/workflows/pr_linting.yml
@@ -1,0 +1,14 @@
+name: PR linter
+
+on: [pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: psf/black@master
+      - uses: isort/isort-action@master
+        with:
+          configuration: " --check-only --diff --profile black"


.github/workflows/pr_linting.yml
```diff
@@ -0,0 +1,10 @@
+name: PR linter
+
+on: [pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: CodeForPoznan/actions/lint-backend@master
```
arturtamborski, line 10:
```suggestion
      - uses: CodeForPoznan/actions/setup-backend@v2
      - uses: CodeForPoznan/actions/lint-backend@v2
```

arturtamborski (changes requested):
This one is now looking pretty good, just missing a few details (see suggestion).

arturtamborski (approved):

Deelite34:
I had a look at this PR.
PR itself is fine, however as I found out, the issue you have lies within [lint backend action in actions repository.](https://github.com/CodeForPoznan/actions/blob/main/lint-backend/action.yml)
It seems that the latest version of black is being used there, despite 22.12.0 version being intended to be used, in result causing this problem.
To fix this inconsistency and use 22.12.0 properly, set version parameter, then I think new release needs to be created, and used in this action.
[source](https://github.com/psf/black/issues/1140#issuecomment-1313674072)

``` 
    - uses: psf/black@22.12.0
      with:
        version: "22.12.0"
        src: backend
```

PR #512, opened 2023-03-08
Changes to project setup instruction

Hybrisen777:
#### Story / Bug id:
[123](paste_your_link_here)

#### Description:
Changes to project setup instruction to accommodate Windows users, added instruction to install missing packages to build a project

#### Migrations:
N/A

N/A

N/A


OtisRed (approved):
It makes perfect sense to add/correct those. Looks good to me!

PR #551, opened 2023-10-11
Vue 3 upgrade

stanislawK:
#### Story / Bug id:
[549](https://github.com/CodeForPoznan/codeforpoznan.pl_v3/issues/549)


frontend/src/plugins/vuetify.js
```diff
@@ -1,11 +1,34 @@
-import '@mdi/font/css/materialdesignicons.css';
-import Vue from 'vue';
-import Vuetify from 'vuetify/lib';
+// eslint-disable-next-line
```
w1stler:
```
/* eslint-disable */
(...)
/* eslint-enable */
```

frontend/src/components/HomePage/Header/HomePageHeader.vue
```diff
@@ -0,0 +1,116 @@
...
+      const el = document.querySelector(element_id);
+
+      el.scrollIntoView({ behavior: 'smooth' });
+    },
+  },
+};
+</script>
+
+<style lang="scss" scoped>
+@import '../../../main.scss';
```
w1stler, line 104: Imports at the top?

w1stler (commented):
Great job, just minor comments for now!

frontend/src/components/HomePage/Header/HomePageHeader.vue
stanislawK: I think that import is strictly related to the <style> section, therefore it is at the top. But I am not sass expert. I found that article, and they suggested a similar way of importing variables in component.

frontend/.eslintrc.json
```diff
@@ -14,7 +14,7 @@
     "prettier/prettier": "error",
     "no-unused-vars": ["warn", { "argsIgnorePattern": "^_" }],
     "no-multi-spaces": ["error", { "ignoreEOLComments": false }],
-    "import/no-extraneous-dependencies": ["error"],
+    "import/no-extraneous-dependencies": ["error", {"devDependencies": true}],
```
OtisRed, line 17: Is it a workaround? Not sure how it works exactly but it seems that `true` is made to ignore this rule [[docs here]](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md).

frontend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:12
+FROM node:19
```
OtisRed, line 1: That's a huge jump :D

frontend/jest.config.js
```diff
@@ -4,5 +4,5 @@ module.exports = {
   testMatch: ['**/?(*.)+(spec).js'],
   collectCoverage: true,
   collectCoverageFrom: ['src/**/*', '!src/assets/**'],
-  transformIgnorePatterns: ['/node_modules/(?!(vue-scrollactive)/)']
+  transformIgnorePatterns: ['/node_modules/(?!(vue-scrollactive)/)'],
```
OtisRed: Is there a reason why we are ignoring this? I'm just curious - if the answer is more than 30sec of writing please ignore this :D

frontend/jest/utils.js
```diff
@@ -1,78 +1,159 @@
-import { mount as nativeMount, createLocalVue } from '@vue/test-utils';
-import Vuetify from 'vuetify';
+import { mount as nativeMount } from '@vue/test-utils';
+import { vi } from 'vitest';
```
OtisRed, line 2: Do I understand correctly that we are nesting Vitest in Jest? If that's correct (just to get up to speed) why we are doing this? Again - if the answer demands too much let's skip it. I'm just fishing if there's something that should be added in the docs to explain how it works.

frontend/src/assets/projects.js
```diff
@@ -1,108 +1,116 @@
+import volontuloImg from '../assets/images/volontulo.png';
+import pahImg from '../assets/images/PAH.png';
+import alinkaImg from '../assets/images/alinka.png';
+import bankEmpatiiImg from '../assets/images/bank_empatii.png';
+import streetmixImg from '../assets/images/streetmix.png';
+import wysadzUliceImg from '../assets/images/wysadz_ulice.png';
+import stronaSpolecznosciImg from '../assets/images/StronaSpolecznosci.png';
+import watchDogImg from '../assets/images/watchdog.png';
```
OtisRed, line 8: That seems so much smarter than what I have done with this before :D

OtisRed (commented):
I went mostly through config files and I have more questions than suggestions. However, this is because I obviously need to learn more to be able to contribute again. The amount of work done is nevertheless notable on the first glance so congratulations on handling something like that. 

I don't want to hold this PR back until I will be able to review this more properly so I suggest we move onward and I'll catch up retroactively. 

However there is one thing that fails at the moment. I wasn't able to log in after populating database. I got POST err everytime I tried one of the users and default password: `pass123`. Before we go on, probably it would be good to check if that's my bad or is it sth with the code on that side of the page. 
Zrzut ekranu 2023-11-22 202601

That being said landing page works like a gold after manual check.

stanislawK:
> I went mostly through config files and I have more questions than suggestions. However, this is because I obviously need to learn more to be able to contribute again. The amount of work done is nevertheless notable on the first glance so congratulations on handling something like that.
> 
> I don't want to hold this PR back until I will be able to review this more properly so I suggest we move onward and I'll catch up retroactively.
> 
> However there is one thing that fails at the moment. I wasn't able to log in after populating database. I got POST err everytime I tried one of the users and default password: `pass123`. Before we go on, probably it would be good to check if that's my bad or is it sth with the code on that side of the page. Zrzut ekranu 2023-11-22 202601
> 
> That being said landing page works like a gold after manual check.

@OtisRed are you able to reach backend endpoints with postman? From the error message, it seems that the backend is not running or it is misconfigured.

frontend/.eslintrc.json
stanislawK: From the documentation you linked, the value `Defaults to true` It means that there is no change in the logic. I don't remember adding that flag manually (although it is very possible that I forgot that I did it :) ) It is possible that it was added automatically when updating some package.

frontend/jest.config.js
stanislawK: I don't know. But because vue-scrollactive supported was dropped, I ended up rewriting scroll logic to vanilla js. In the newest PR I removed vue-scrollactive, to avoid confusion. Thank you for pointing that out.

frontend/jest/utils.js
stanislawK: We are not nesting it, but unfortunately we are using both jest and vitest for now. At some point we need to reconfigure eslint to use vitest, and move all subpackages that are dependent on jest to something different (and most likely reconfigure that too).
Right now we run unittest with vitest, but eslint is configure with jest, as well as some subpackages.
Removing jest is something we should most likely do, but at lest for me it is not a simple task.

PR #558, opened 2026-06-16
Fix backend tech stack factory flake

OtisRed:
## Summary
- make TechStackFactory generate deterministic unique technology values
- add regression tests for batch creation and duplicate fallback

## Verification
- python -m py_compile backend/factories.py backend/tests/test_factories.py
- Full local pytest not run: pipenv is not installed in this local environment. GitHub Actions should verify the backend workflow.

### DIFF
--- a/backend/factories.py
+++ b/backend/factories.py
@@ -1,4 +1,5 @@
 from random import randint
+from uuid import uuid4
 
 import factory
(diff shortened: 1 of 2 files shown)


OtisRed:
Yolo

